PR #5, opened 2026-02-16
enclave support

aruokhai:
## Summary

- Add a complete AWS Nitro Enclave deployment scaffold for running the introspector signing oracle inside an isolated enclave, ensuring the signing key is never exposed to the host instance
- Includes enclave configuration, gvproxy networking, systemd service units, EC2 cloud-init user data, and the enclave entrypoint scripts
- Based on the [introspector-enclave](https://github.com/ArkLabsHQ/introspector-enclave) reference architecture

### Configuration (`enclave/enclave.yaml`)
Declarative enclave deployment config specifying:
- Target AWS region, account, and instance type (`m6i.xlarge`)
- Nix-based reproducible build pinned to a specific introspector commit (`dcec46c`)
- Application port (`7074`) and secret references (signing key via KMS)

### Networking (`enclave/gvproxy/`)
- **Dockerfile**: Multi-stage build producing a minimal Alpine container with [gvproxy](https://github.com/containers/gvisor-tap-vsock) v0.7.4 for vsock-based outbound networking
- **start.sh**: Launches gvproxy and configures port forwarding from the host into the enclave's virtual IP (`192.168.127.2`)

### Enclave entrypoint (`enclave/start.sh`)
Orchestrates the enclave boot sequence:
1. Starts viproxy for IMDS access (credential fetching inside the enclave)
2. Configures DNS via the gvproxy gateway (`192.168.127.1`)
3. Launches [nitriding](https://github.com/brave/nitriding-daemon) for TLS termination, attestation, and app supervision

### Host-side init (`enclave/scripts/enclave_init.sh`)
Starts the Nitro Enclave via `nitro-cli run-enclave` with configurable CPU, memory, and CID. Polls until the enclave exits, designed to run under systemd with `Restart=always`.

### Systemd services (`enclave/systemd/`)
| Service | Purpose |
|---------|---------|
| `enclave-watchdog.service` | Starts and auto-restarts the Nitro Enclave |
| `enclave-imds-proxy.service` | Forwards IMDS requests into the enclave via vsock |
| `gvproxy.service` | Runs the gvproxy Docker container for outbound networking (ports 443, 7073, 9090) |

### EC2 user data (`enclave/user_data/user_data`)
Cloud-init script that provisions a fresh EC2 instance:
- Installs Nitro Enclaves CLI and allocates enclave resources (6 GiB RAM, 2 vCPUs)
- Configures the vsock proxy allowlist (KMS, SSM, STS, IMDS)
- Downloads the pre-built EIF from S3 and the gvproxy image from ECR
- Installs and enables all systemd services
- Writes enclave environment variables to `/etc/environment`


louisinger (changes requested):
we need to document this flow in a new enclave/README file describing each file, link to potential documentation etc...

.github/workflows/build-eif.yml
```diff
@@ -0,0 +1,108 @@
+name: Build EIF
+
+on:
+  push:
+    branches: [master, main]
```
louisinger:
```suggestion
    branches: [master]
```

.github/workflows/build-eif.yml
```diff
@@ -0,0 +1,108 @@
+name: Build EIF
+
+on:
+  push:
```
louisinger: should not build on tag (release) instead of every push ?

aruokhai:
@louisinger  This env seems to be problematic **INTROSPECTOR_ARKD_URL**, because we need to have all environment variables baked into the EIF. I reckon that it will be different for mutiny and mainnet.

louisinger:
> @louisinger This env seems to be problematic **INTROSPECTOR_ARKD_URL**, because we need to have all environment variables baked into the EIF. I reckon that it will be different for mutiny and mainnet.

I need more context to understand. What is EIF and why it's a problem if mutiny/mainnet env is different ?
Also the changes made in this PR seem to be infrastructure related. 

do we need/want them to be public in the repository ?

aruokhai:
> > @louisinger This env seems to be problematic **INTROSPECTOR_ARKD_URL**, because we need to have all environment variables baked into the EIF. I reckon that it will be different for mutiny and mainnet.
> 
> I need more context to understand. What is EIF and why it's a problem if mutiny/mainnet env is different ? Also the changes made in this PR seem to be infrastructure related.
> 
> do we need/want them to be public in the repository?

EIF is similar to a Docker image, but for an enclave. The yaml file is used for building the enclave and deployment. For now, we bake the env variables into the EIF file.  I reckon that I can include a runtime configuration functionality, so deployment can configure such environment variables. 
The overall idea is for the infra to ingest the EIF file and use it during deployment (similar to how an infra will use a Docker image), which means that the image file needs to be built by the introspector repo.

PR #10, opened 2026-02-28
feat: add OP_TXID opcode

### LINKED ISSUE #7: OP_TXID
Need a way to push the tx's hash to the stack

Kukks:
## Summary
- Add `OP_TXID` opcode at slot `0xf3` (243), replacing `OP_UNKNOWN243`
- The opcode pushes the current transaction's hash (txid) onto the stack as a 32-byte array
- Follows the same pattern as existing introspection opcodes like `OP_INSPECTVERSION`

## Changes
- **pkg/arkade/opcode.go**: Define `OP_TXID` constant, register in `opcodeArray`, and implement `opcodeTxId` handler
- **pkg/arkade/opcode_test.go**: Update `TestOpcodeDisasm` expected strings and unknown opcode ranges
- **pkg/arkade/engine_test.go**: Add three test cases:
  - `OP_TXID`: Verify the pushed hash matches the expected transaction hash
  - `OP_TXID_LENGTH`: Verify the pushed value is exactly 32 bytes
  - `OP_TXID_WRONG_HASH`: Verify script fails when comparing against an incorrect hash

## Test plan
- [x] All existing tests pass
- [x] New OP_TXID tests pass (positive, length check, and negative case)
- [x] `go fmt` applied

Closes #7


louisinger (changes requested):
let's update README table too

PR #30, opened 2026-03-16
Support every type of tapscript closure

louisinger:
Fix an old TODO preventing any type of tapscript to be used with the introspector.

@chris-ricketts @Kukks please review


louisinger:
> @louisinger looks good to me. I suppose it would be good to have an integration test case for each type of closure, although that could easily be a separate issue.

@chris-ricketts I'm adding unit tests with psbt fixtures to `ReadArkadeScript` should be ok-ish now ? I'll open issue for integrations

PR #35, opened 2026-03-17
Fuzz: DeserializeIntrospectorPacket

chris-ricketts:
## Changes:
- Add fuzz coverage for DeserializeIntrospectorPacket, seeded from existing fixture encodings plus a regression corpus input.
- Extend invalid fixture vectors with entry-count, script-length, and witness-length overflow cases.
- Bound introspector packet decode sizes to fail with errors instead of panic on malformed inputs.

## Notes:
- This PR is blocked on upstream ark-lib witness decode hardening.
- Upstream PR: https://github.com/arkade-os/arkd/pull/974
- CI is expected to fail until pkg/arkade/go.mod is bumped to include that upstream fix.


chris-ricketts:
Marked as draft until https://github.com/arkade-os/arkd/pull/974 is merged and we can bump ark-lib dependency - CI expected to fail on `test / unit` until then.

pkg/arkade/introspector_packet.go
```diff
@@ -133,6 +148,9 @@ func DeserializeIntrospectorPacket(data []byte) (IntrospectorPacket, error) {
 		if err != nil {
 			return nil, fmt.Errorf("failed to read witness_len for entry %d: %w", i, err)
 		}
+		if witnessLen > MaxWitnessLength {
+			return nil, fmt.Errorf("max introspector witness length exceeded, max=%d got=%d", MaxWitnessLength, witnessLen)
+		}
 		witnessBytes := make([]byte, witnessLen)
```
chris-ricketts: If we did not want to set an arbitrary limit on total witness length, we could also do:

```suggestion
		remainingBytes := r.Len()
		if witnessLen > uint64(remainingBytes) {
			return nil, fmt.Errorf("witness length value exceeds remaining bytes, actual=%d got=%d", remainingBytes, witnessLen)
		}
		witnessBytes := make([]byte, witnessLen)
```

chris-ricketts:
@louisinger / @Kukks please review

PR #44, opened 2026-03-31
Test/Fix: Negative index rejection for input/output introspection opcodes

chris-ricketts:
Changes:

- 3169ef6: add regression coverage in `pkg/arkade/engine_test.go` for negative indices across all six input/output introspection opcodes, asserting explicit error strings for invalid negative index handling.

- e0c651e: fix `pkg/arkade/opcode.go` by adding `< 0` guards before upper-bound checks for introspection opcodes, returning `cannot be negative` for negative indices and preserving out of range for max-bound errors.


louisinger (approved):
ACK, up to you to address my nit comment or not. feel free to squash and merge it

PR #49, opened 2026-04-01
Asset introspection pushes LE64 instead of scriptNum

msinkec:
This PR fixes asset amount handling in asset introspection opcodes by moving amount values from `scriptNum` to 8-byte little-endian (`LE64`) stack values, and updates lookup semantics to a flag+value pattern.

### Why

Asset amounts are defined as `uint64`, but previously they were pushed as `scriptNum` (`int64`) and later parsed with a 4-byte `MakeScriptNum` limit via `PopInt()`. That created correctness bugs for larger amounts:

- values above `2^63-1` could wrap on cast to `int64`
- values above ~`2^31-1` could fail with `ErrNumberTooBig` when re-read as script numbers

### What changed

- Added LE64 push helper in `asset_opcodes.go` and switched amount pushes to LE64 for:
  - `OP_INSPECTASSETGROUP` (input/output amount fields)
  - `OP_INSPECTASSETGROUPSUM`
  - `OP_INSPECTOUTASSETAT`
  - `OP_INSPECTINASSETAT`
- Updated lookup opcode semantics:
  - `OP_INSPECTOUTASSETLOOKUP`
  - `OP_INSPECTINASSETLOOKUP`
  - **Found:** push `1` then `<amount LE64>`
  - **Not found:** push `0` only
- Updated tests to compare amounts as LE64 and assert new lookup stack layout.
- Updated integration test script helper in `test/utils_test.go` to compare asset sums as LE64.

### Script ergonomics improvement (Issue #40)

Using asset lookup amounts is now simpler and more consistent with 64-bit ops, matching the direction in [Issue #40](https://github.com/ArkLabsHQ/introspector/issues/40):

- no `-1` sentinel handling boilerplate
- no `OP_SCRIPTNUMTOLE64` conversion step before `OP_ADD64` / `OP_MUL64` / comparisons
- straightforward found-check via top-stack flag (`OP_VERIFY`), with amount already in LE64

### Edge cases covered

Added explicit tests for large amounts and arithmetic/comparison interoperability:

- large amount reads from group/input/output paths
- large group sums (e.g. 10B)
- direct use with `OP_ADD64`
- direct use with `OP_GREATERTHANOREQUAL64`
- lookup found path with large amount using flag+LE64

### Breaking change / migration note

This is a **breaking stack-shape/encoding change** for scripts using asset amount outputs:

- Amounts are now LE64 byte arrays, not `scriptNum`.
- Lookup opcodes now return `flag + amount` (found) or `flag` only (not found), replacing the old `-1` sentinel behavior.

### DIFF
--- a/pkg/arkade/asset_opcodes.go
+++ b/pkg/arkade/asset_opcodes.go
@@ -10,15 +10,6 @@ import (
 	"github.com/btcsuite/btcd/txscript"
 )
 
-// pushAmountLE64 encodes a uint64 as 8-byte little-endian and pushes it onto
-// the data stack. This is consistent with how OP_INSPECTOUTPUTVALUE pushes
-// satoshi values, allowing scripts to use 64-bit arithmetic ops directly.
-func pushAmountLE64(vm *Engine, amount uint64) {
-	buf := make([]byte, 8)
-	binary.LittleEndian.PutUint64(buf, amount)
-	vm.dstack.PushByteArray(buf)
-}
-
 // opcodeInspectNumAssetGroups pushes the total number of asset groups in the packet onto the stack.
 func opcodeInspectNumAssetGroups(op *opcode, data []byte, vm *Engine) error {
 	if vm.assetPacket == nil {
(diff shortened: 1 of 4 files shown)


pkg/arkade/asset_opcodes.go
```diff
@@ -386,7 +396,8 @@ func opcodeInspectOutAssetAt(op *opcode, data []byte, vm *Engine) error {
 }
 
 // opcodeInspectOutAssetLookup pops gidx, txid, and output index o, then looks up the asset amount.
-// Pushes the amount if found, or -1 if not found.
+// Found:     pushes 1 (success flag) then the amount as 8-byte LE64.
+// Not found: pushes 0 (failure flag only — no amount on stack).
```
louisinger:
```suggestion
// Not found: pushes -1 (failure flag only — no amount on stack).
```
why not -1 to be consistent with other opcodes ?

louisinger (commented):
The problem with returning 0 for "not found" is that the script author has to remember two patterns, because OP_FINDASSETGROUPBYASSETID still uses -1. Also, we could have asset outputs with an amount of 0 allowed... I'd keep -1 if possible.

msinkec:
The the reason I don't really like -1 is that OP_IF doesn't accept it (not minimal encoding) and that OP_VERIFY doesn't fail on -1. Assets with amount of 0 aren't a problem, since flag value is unrelated to the amount value.

Issue #52, opened 2026-04-05
INSPECTINPUTSCRIPTPUBKEY should return prevout ark script instead of checkpoint scriptpubkey


louisinger:
While executing `INSPECTINPUTSCRIPTPUBKEY` opcode on offchain transaction, the engine should return the script public key of the previous ark tx (input of the checkpoints). for other type of transactions, the current behavior should be unchanged.

PR #55, opened 2026-04-08
Implement OP_INSPECTPACKET and OP_INSPECTINPUTPACKET.

### LINKED ISSUE #50: add INSPECTINPUTPACKET
`<packet_type> <input_index> INSPECTINPUTPACKET` returns the packet with `packet_type` if present in prevout transaction at `input_index`. it should handle the case where the packet is not found (push 0/-1 on the stack?)

the prevout ark transaction (not the checkpoints) must be set by the user in a dedicated [unknown ark psbt field](https://github.com/arkade-os/arkd/blob/d372608bf69c541bf29d6f5cef6b627e002eac1c/pkg/ark-lib/txutils/psbt_fields.go#L27-L36). We don't want to use NonWitnessUtxo because it's non standard.

### LINKED ISSUE #51: add INSPECTPACKET
`<packet_type> INSPECTPACKET` should return the `packet_type` of the current transaction. it should handle the case where the packet is not found (0/-1 on the stack ?)

msinkec:
- **`OP_INSPECTPACKET`** (0xf4): Pops a packet type, returns the raw contents of the matching TLV packet from the current transaction's ark extension. Pushes `content, 1` if found; `<empty>, 0` if not.
- **`OP_INSPECTINPUTPACKET`** (0xf5): Pops an input index and packet type, looks up the previous ark transaction for that input, and returns the matching packet contents. Same found/not-found semantics.
- Previous ark transactions are provided to the engine via `SetPrevArkTxs(map[int]*wire.MsgTx)`, intended to be populated from a dedicated ark PSBT unknown field (`[0xde]["prevarktx"]`), not `NonWitnessUtxo`.
- Packet content exceeding `MaxScriptElementSize` (520 bytes) is rejected.
- Stack depth is kept uniform (always 2 elements: content + flag), matching the asset introspection convention.

### Test plan
- [x] Found / not-found for both opcodes
- [x] Multiple packet types in one extension
- [x] No extension, no prev ark tx, nil prev ark txs map
- [x] Negative / out-of-range input index
- [x] Empty stack
- [x] Large packet content (>520 bytes) rejected

Closes #50
Closes #51


pkg/arkade/opcode.go
```diff
@@ -3231,6 +3232,128 @@ func opcodeTxId(op *opcode, data []byte, vm *Engine) error {
...
+		vm.dstack.PushByteArray(nil)
+		vm.dstack.PushInt(0)
+		return nil
+	}
+
+	if len(content) > txscript.MaxScriptElementSize {
+		str := fmt.Sprintf("packet content size %d exceeds max allowed size %d",
+			len(content), txscript.MaxScriptElementSize)
+		return scriptError(txscript.ErrElementTooBig, str)
+	}
```
louisinger, line 3348: open question, not only for this opcode : do we really need that element size limit ?

pkg/arkade/engine.go
```diff
@@ -191,6 +192,15 @@ func (vm *Engine) SetIntrospectorPacket(packet IntrospectorPacket) {
 	vm.introspectorPacket = packet
 }
 
+// SetPrevArkTxs sets the previous ark transactions for input packet
+// introspection. The map key is the input index of the currently executing
+// transaction; the value is the full previous ark transaction that the
+// input references. These are provided via a dedicated ark PSBT unknown
+// field (key 0xde "prevarktx") and are NOT the same as NonWitnessUtxo.
+func (vm *Engine) SetPrevArkTxs(txs map[int]*wire.MsgTx) {
```
louisinger:
```suggestion
func (vm *Engine) SetPrevoutTxs(txs map[int]*wire.MsgTx) {
```
I'd not name it Prev"Ark"Txs because in case of intents, you could also spend onchain inputs with packet/state. For instnace boarding and/or previously unrolled vtxo tx.

louisinger:
It also need the implementation of the application layer before merge (set the prevout tx + validation)

pkg/arkade/opcode.go
msinkec: I'd keep it for now, we could raise it later. As for wether we need the check here, yes it's standard practice to do this. We're also missing this check in some other opcodes like OP_CAT.

pkg/arkade/opcode.go
```diff
@@ -3231,6 +3232,128 @@ func opcodeTxId(op *opcode, data []byte, vm *Engine) error {
...
+		return err
+	}
+
+	if packetType < 0 || packetType > 255 {
+		return scriptError(txscript.ErrInvalidStackOperation, "packet type out of range")
+	}
+
+	content, err := findPacketByType(&vm.tx, uint8(packetType))
+	if err != nil {
+		return scriptError(txscript.ErrInvalidStackOperation, err.Error())
```
chris-ricketts, line 3277: Nit: I don't think `ErrInvalidStackOperation` is the correct error variant here based on the doc comment:

```go
// ErrInvalidStackOperation is returned when a stack operation is
// attempted with a number that is invalid for the current stack size.
ErrInvalidStackOperation
```

However, there does not seem to be a better option in btcd.

It might be worth adding our own extension to `txscript.ErrorCode` in `pkg/arkade/error.go`, like:

```go
const (
    // ErrMalformedArkadeExtension is returned when a transaction contains an
    // Arkade extension whose serialized payload is invalid and cannot be parsed.
	ErrMalformedArkadeExtension txscript.ErrorCode = 10_000 + iota
	/// ...
)
```

Same for:
https://github.com/ArkLabsHQ/introspector/blob/0b2971b7e5296e77ee354100d8d9416eb68bdc46/pkg/arkade/opcode.go#L3335-L3338

pkg/arkade/opcode.go
```diff
@@ -3231,6 +3232,128 @@ func opcodeTxId(op *opcode, data []byte, vm *Engine) error {
...
 }
 
+// findPacketByType scans a transaction for an ark extension packet of the given
+// type and returns its serialized content. Returns nil if no extension is
+// present or the requested type is not found.
+func findPacketByType(tx *wire.MsgTx, packetType uint8) ([]byte, error) {
+	ext, err := extension.NewExtensionFromTx(tx)
+	if err != nil {
+		return nil, nil // no extension
+	}
```
chris-ricketts, line 3246: `err` is not always `ErrExtensionNotFound` as `extension.NewExtensionFromTx` calls `extension.NewExtensionFromBytes` which returns parsing errors:

https://github.com/arkade-os/arkd/blob/d372608bf69c541bf29d6f5cef6b627e002eac1c/pkg/ark-lib/extension/extension.go#L131-L190

```suggestion
	ext, err := extension.NewExtensionFromTx(tx)
	if err != nil {
		if errors.Is(err, extension.ErrExtensionNotFound) {
			return nil, nil
		}
		return nil, fmt.Errorf("failed to parse extension: %w", err)
	}
```

pkg/arkade/opcode.go
```diff
@@ -3231,6 +3232,128 @@ func opcodeTxId(op *opcode, data []byte, vm *Engine) error {
...
+// Not found: pushes empty byte array, then 0.
+// Stack transformation: [... packet_type] -> [... content 1] or [... <empty> 0]
+func opcodeInspectPacket(op *opcode, data []byte, vm *Engine) error {
+	packetType, err := vm.dstack.PopInt()
+	if err != nil {
+		return err
+	}
+
+	if packetType < 0 || packetType > 255 {
+		return scriptError(txscript.ErrInvalidStackOperation, "packet type out of range")
```
chris-ricketts, line 3272: Nit: similar to https://github.com/ArkLabsHQ/introspector/pull/55/changes#r3050098534

`ErrInvalidIndex` kind of fits better but is not ideal. 

Same in:
https://github.com/ArkLabsHQ/introspector/blob/0b2971b7e5296e77ee354100d8d9416eb68bdc46/pkg/arkade/opcode.go#L3318-L3320

msinkec:
Implemented end-to-end PSBT plumbing for previous input transactions, now exposed as a `prevouttx` Ark PSBT unknown field. The change adds encoding/decoding for the per-input transaction payload, validates that any provided transaction matches the input’s referenced prevout hash, and wires the decoded map into Arkade execution so `OP_INSPECTINPUTPACKET` can inspect packets from the spent transaction.

The application layer now loads these fields before script execution in both `SubmitTx` and `SubmitIntent`, while preserving the existing checkpoint-signing flow. Added focused tests for field round-tripping, duplicate-field rejection, and prevout-hash mismatch handling.

pkg/arkade/opcode.go
msinkec: This would also be the case for many other opcodes then. Probably better to move to a separate issue and address them all at once.

chris-ricketts: Created: https://github.com/ArkLabsHQ/introspector/issues/57

pkg/arkade/opcode.go
chris-ricketts: Created: https://github.com/ArkLabsHQ/introspector/issues/57

chris-ricketts (approved):

pkg/arkade/psbt_fields.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+		if !ok {
+			continue
+		}
+
+		fieldsFound = append(fieldsFound, value)
+	}
+
+	return fieldsFound, nil
+}
```
louisinger: can't we import from ark-lib ? I understand the "double pointer" thing but maybe the psbt field itself should by wire.MsgTx ? or bytes ? or whatever allowing to use the ark-lib coders ?

pkg/arkade/psbt_fields.go
```diff
@@ -0,0 +1,146 @@
...
+
+func PrevoutTxsFromPSBT(ptx *psbt.Packet) (map[int]*wire.MsgTx, error) {
```
louisinger: that function seems to be only server side validation. let's move it to internal/application pkg instead ?

pkg/arkade/psbt_fields.go
```diff
@@ -0,0 +1,146 @@
...
+	if len(ptx.Inputs) != len(ptx.UnsignedTx.TxIn) {
+		return nil, fmt.Errorf("malformed psbt")
+	}
+
+	prevoutTxs := make(map[int]*wire.MsgTx)
+
+	for inputIndex := range ptx.Inputs {
```
louisinger: should be optional to set the prevout and throw in the script at runtime only if the prevout tx is really needed. it doesn't introduce a breaking change + it makes tx lighter if your script is not using any inspect packet opcodes.

louisinger (changes requested):

pkg/arkade/psbt_fields.go
msinkec: Already optional.

- PrevoutTxsFromPSBT iterates inputs and simply skips inputs with no prevarktx field.
- if no fields are present at all, it returns an empty map, not an error.
- SubmitTx / SubmitIntent always call WithPrevoutTxs(...), but passing an empty map is harmless.
- in the VM, OP_INSPECTINPUTPACKET takes the not-found path if:
      - the map is nil, or
      - the input index is absent from the map

internal/application/prevout.go
```diff
@@ -0,0 +1,50 @@
...
+
+		if len(fields) == 0 {
+			continue
+		}
+		if len(fields) > 1 {
+			return nil, fmt.Errorf("multiple prevout tx fields found for input %d", inputIndex)
+		}
+
+		prevTx := fields[0]
+		expectedHash := ptx.UnsignedTx.TxIn[inputIndex].PreviousOutPoint.Hash
```
louisinger: I think this won't work for offchain ark tx. for intent it's ok but offchain transaction input outpoints are checkpoints so it will never match the prevout tx hash. 
we may need a `validatePrevoutTxsForIntent(intentPtx, prevouts)` and `validatePrevoutTxsForArkTx(arkPtx, prevouts, checkpoints)`

internal/application/prevout.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/ArkLabsHQ/introspector/pkg/arkade"
+	"github.com/arkade-os/arkd/pkg/ark-lib/txutils"
+	"github.com/btcsuite/btcd/btcutil/psbt"
+	"github.com/btcsuite/btcd/wire"
+)
+
+// PrevoutTxsFromPSBT extracts per-input prevout transactions from Ark PSBT
+// unknown fields and validates that each provided transaction matches the
+// prevout hash referenced by the corresponding input.
+func PrevoutTxsFromPSBT(ptx *psbt.Packet) (map[int]*wire.MsgTx, error) {
```
louisinger:
```suggestion
func prevoutTxsFromPSBT(ptx *psbt.Packet) (map[int]*wire.MsgTx, error) {
```
should not be exported. also let's split into 2 different functions : readPrevoutTxs(ptx) and validatePrevoutTx(ptx, prevouts). this way the protocol validation logic is in a dedicated function.

louisinger (changes requested):

pkg/arkade/psbt_fields_test.go
```diff
@@ -0,0 +1,65 @@
+package arkade
```
louisinger, line 1: this file should follow the same pattern of `pkg/arkade/tesdata` tests (JSON fixtures, invalid/valid cases) because that's something we want to port to other languages

internal/application/prevout.go
msinkec: Pushed a commit that handles this now.

msinkec:
Also found and fixed an issue with the ommition of prev tx data.

**Fix** (`pkg/arkade/opcode.go:3326-3337`): `OP_INSPECTINPUTPACKET` now **errors** when the prevout tx is missing, instead of silently returning "not found." This makes the distinction:

| Situation | Before | After |
|---|---|---|
| Prevout txs not provided at all | `(nil, 0)` — soft "not found" | **Script error** |
| Prevout tx missing for specific input | `(nil, 0)` — soft "not found" | **Script error** |
| Prevout tx provided, packet type not in it | `(nil, 0)` | `(nil, 0)` — unchanged |
| Prevout tx provided, packet found | `(content, 1)` | `(content, 1)` — unchanged |

This way scripts cannot be tricked regardless of how their branching logic is written — the omission attack causes a hard failure at the opcode level before any script branch is taken.

pkg/arkade/engine.go
```diff
@@ -191,6 +192,15 @@ func (vm *Engine) SetIntrospectorPacket(packet IntrospectorPacket) {
 	vm.introspectorPacket = packet
 }
 
+// SetPrevoutTxs sets the previous ark transactions for input packet
+// introspection. The map key is the input index of the currently executing
+// transaction; the value is the full previous ark transaction that the
+// input references. These are provided via a dedicated ark PSBT unknown
+// field (key 0xde "prevarktx") and are NOT the same as NonWitnessUtxo.
+func (vm *Engine) SetPrevoutTxs(txs map[int]*wire.MsgTx) {
```
louisinger:
```suggestion
func (vm *Engine) SetArkPrevoutTxs(txs map[int]*wire.MsgTx) {
```

pkg/arkade/engine.go
```diff
@@ -191,6 +192,15 @@ func (vm *Engine) SetIntrospectorPacket(packet IntrospectorPacket) {
 	vm.introspectorPacket = packet
 }
 
+// SetPrevoutTxs sets the previous ark transactions for input packet
+// introspection. The map key is the input index of the currently executing
+// transaction; the value is the full previous ark transaction that the
+// input references. These are provided via a dedicated ark PSBT unknown
+// field (key 0xde "prevarktx") and are NOT the same as NonWitnessUtxo.
```
louisinger:
```suggestion
// input references. 
```
in case of unrolled vtxo, we won't be able to get them from the psbt fields.

pkg/arkade/engine.go
louisinger: also no need to be exported, it seems only used within the package ? maybe the option func is enough, no need for this extra setter ?

msinkec:
Added an e2e counter-contract test covering the new packet introspection opcodes.

  Flow:
  - fund Alice and send to a staging introspector-controlled VTXO
  - deploy the counter with packet value `0` using `OP_INSPECTPACKET`
  - try an invalid unlock that keeps value `0` and assert introspector rejects it
  - unlock with value `1`, using `OP_INSPECTINPUTPACKET` to read the previous packet
  - finalize through introspector + arkd, including checkpoint signatures
  - spend the checkpointed counter output again with value `2`
  - assert value, amount, and scriptPubKey are propagated forward

pkg/arkade/engine.go
msinkec: Removed the setter now.

pkg/arkade/engine.go
```diff
@@ -103,14 +103,15 @@ type Engine struct {
...
-	assetPacket    asset.Packet
+	tx                 wire.MsgTx
+	txIdx              int
+	version            uint16
+	sigCache           *txscript.SigCache
+	hashCache          *txscript.TxSigHashes
+	prevOutFetcher     txscript.PrevOutputFetcher
+	assetPacket        asset.Packet
 	introspectorPacket IntrospectorPacket
+	prevoutTxs         map[int]*wire.MsgTx
```
chris-ricketts: Thinking about how this PR interacts with the implementation for #52.

I think `prevoutTxs` may need to carry slightly more information than `map[int]*wire.MsgTx`.
That shape is sufficient for opcodes like `OP_INSPECTINPUTPACKET`, which only need the previous Ark tx itself. But for `OP_INSPECTINPUTSCRIPTPUBKEY` in the offchain/checkpoint case, we need two pieces of information per input:
1. the previous Ark tx
2. which output of that tx is the one being continued

  For an offchain Ark tx input, those are not derivable from the current input alone:
  - the Ark tx input directly spends the checkpoint output
  - but the scriptPubKey we want is on the previous Ark tx output referenced by the checkpoint input's prevout

So for this case, `map[int]*wire.MsgTx` gives us the correct tx, but not the correct output index within that tx.

Would it make sense to represent this as something closer to `map[int]PrevoutTxRef`, where:

```go
type PrevoutTxRef struct {
    Tx          *wire.MsgTx
    OutputIndex uint32
}
```
That would let:
- packet inspection keep using `ref.Tx`
- scriptPubKey inspection use `ref.Tx.TxOut[ref.OutputIndex]`

I think this may be cleaner than keeping prevoutTxs as just the tx and introducing a second parallel `map[inputIndex]uint32` when implementing the fix for #52.

Just looking for feedback - I can make this change in the PR for #52 once this is merged.

louisinger: nice catch, I think we should follow the same pattern of the existing prevoutFetcher (the "normal" prevouts) = the engine takes an interface, then application layer implements it. so in this PR the interface will be : 

```
type ArkPrevOutFetcher interface {
   FetchPrevOutArkTx(wire.Outpoint) *wire.MsgTx
}
```

then we add a new `FetchPrevOutScriptPubKey(wire.Outpoint)` when we implement #52. 

the interface IMO has several advantages :  
* move complex implementation logic outside of the engine 
* application layer can implement several types of prevout fetcher (read from PSBT, use indexer etc..., intent etc...), without touching the -sensitive- engine code.
* easier for testing ?

cc @msinkec @chris-ricketts

test/utils_test.go
```diff
@@ -718,6 +718,79 @@ func executeArkadeScripts(t *testing.T, ptx *psbt.Packet, signerPublicKey *btcec
...
+		for inputIndex, prevTx := range prevoutTxs {
+			checkpointTxid := ptx.UnsignedTx.TxIn[inputIndex].PreviousOutPoint.Hash.String()
+			checkpoint, ok := checkpointsByTxid[checkpointTxid]
+			if !ok {
+				return fmt.Errorf("checkpoint not found for input %d", inputIndex)
+			}
+			if len(checkpoint.UnsignedTx.TxIn) == 0 {
+				return fmt.Errorf("checkpoint has no inputs for input %d", inputIndex)
+			}
+
```
louisinger: do we really need the validation logic for testing, can't we admit they are valid ?

internal/application/intent.go
```diff
@@ -53,7 +58,12 @@ func (s *service) SubmitIntent(ctx context.Context, intent Intent) (*psbt.Packet
 			continue
 		}
 
-		if err := script.Execute(ptx.UnsignedTx, prevoutFetcher, inputIndex); err != nil {
+		if err := script.Execute(
+			ptx.UnsignedTx,
+			prevoutFetcher,
+			inputIndex,
+			arkade.WithPrevoutTxs(prevoutTxs),
+		); err != nil {
```
louisinger, line 60:
```suggestion
		if err := script.Execute(
			ptx.UnsignedTx,
			prevoutFetcher,
			arkPrevoutFetcher,
			inputIndex,
		); err != nil {
```
followup suggstions of chris' comment : let's remove the option pattern and pass an interface implementation (similar to what's done in original prevoutFetcher)

chris-ricketts: Makes sense, then the `opcodeInspectInputScriptPubkey` impl doesn't have to worry about falling back to the normal `prevoutFetcher` in the case where it can't find it in the (potentially empty) `prevoutTx`

test/utils_test.go
chris-ricketts: Good point, that checkpoints param is not needed at all. Fixed in b754f8c8a5a4d4eedc203d0d854db32ae2cf2b1f

pkg/arkade/engine.go
chris-ricketts: We need to ensure that the `btcd` `PrevOutputFetcher` is still implemented the same (returning the checkpoint outputs) as this is used in signature verification:

https://github.com/ArkLabsHQ/introspector/blob/b754f8c8a5a4d4eedc203d0d854db32ae2cf2b1f/pkg/arkade/script.go#L121

https://github.com/ArkLabsHQ/introspector/blob/b754f8c8a5a4d4eedc203d0d854db32ae2cf2b1f/pkg/arkade/sigvalidate.go#L170-L172

https://github.com/ArkLabsHQ/introspector/blob/b754f8c8a5a4d4eedc203d0d854db32ae2cf2b1f/pkg/arkade/sigvalidate.go#L240-L242

pkg/arkade/script.go
```diff
@@ -34,10 +34,25 @@ func WithDebugCallback(callback func(*StepInfo, *Engine) error) ExecuteOption {
...
-		engine.prevoutTxs = prevoutTxs
+// ArkPrevOutFetcher looks up previous ark transactions by outpoint.
+type ArkPrevOutFetcher interface {
+	FetchPrevOutArkTx(wire.OutPoint) *wire.MsgTx
+}
+
+// MapArkPrevOutFetcher is a map-based implementation of ArkPrevOutFetcher.
+type MapArkPrevOutFetcher struct {
+	m map[wire.OutPoint]*wire.MsgTx
+}
```
chris-ricketts: We could extend the `txscript.PrevOutputFetcher` interface instead of adding another separate one which results in 2 pretty similar parameters/fields.

```suggestion
type ArkPrevOutFetcher interface {
    // extends
	txscript.PrevOutputFetcher

	FetchPrevOutArkTx(wire.OutPoint) *wire.MsgTx
}

// MapArkPrevOutFetcher is a map-based implementation of ArkPrevOutFetcher.
type MapArkPrevOutFetcher struct {
    // Checkpoint or Previous ArkTx depending on SubmitTx or SubmitIntent resp.
	prevouts   map[wire.OutPoint]*wire.TxOut
	// Always Previous ArkTx
	prevoutTxs map[wire.OutPoint]*wire.MsgTx
}
```

msinkec: Yea, that's proably a good idea. I've unified the interfaces now. I feel like the sematic is still sort of awkward but don't really have an good ideas to improve. I guess it's good for now. I've also added some comments to make it clearer.

pkg/arkade/psbt_fields_test.go
louisinger: cc @msinkec 
we need this to port the change to JS

pkg/arkade/script.go
louisinger: concept ACK but the implementation should be in internal/application/prevout.go **unexported**. only the interface lives here in the engine pkg.

internal/application/prevout_test.go
```diff
@@ -0,0 +1,163 @@
+package application
```
louisinger, line 1: let's also rely on JSON psbt fixtures here because : 
* it will make the file easier to read / smaller
* if we find a buggish psbt in the future, we can just add it to fixture without having to "rebuild" it with go code

JSON should be : 

```
{
   "valid": [
         {
             "name": "intent psbt",
              "psbt": ".......",
              "checkpoints": []
         },
         ...
    ] ,
    "invalid": [
         {
              "name": "duplicate prevout psbt fields",
              "psbt": "......",
              "expectedErr": "...."
     ] 
}
```

then the test : 

```
func TestArkPrevOutFetcher(t *testing.T) {
      t.Run("valid", .....)
      t.Run("invalid", .....)
} 
```

louisinger (changes requested):

pkg/arkade/script.go
msinkec: Ok but in this case we need re-implementations of the fetcher in each package for the tests.

internal/application/prevout_test.go
msinkec: Added fixtures and support code.

pkg/arkade/script.go
msinkec: Done in [9c0f853](https://github.com/ArkLabsHQ/introspector/pull/55/commits/9c0f853328e08b5c238c5786872a22b4a243b0c4)

chris-ricketts:
Rebased onto `master` to get #62 and clear Trivy CI failure

internal/application/testdata/ark_prevout_fetcher.json
```diff
@@ -0,0 +1,34 @@
+{
```
louisinger, line 1: remove this please, we must have a single `ark_prevout_fetcher.json` with all data (psbt/checkpoints in that case). as reference, check ReadArkadeScript unit tests fixtures : https://github.com/ArkLabsHQ/introspector/blob/master/pkg/arkade/testdata/read_arkade_script.json 

or in arkd for intent psbt : https://github.com/arkade-os/arkd/blob/master/pkg/ark-lib/intent/testdata/proof_fixtures.json

single JSON with valid and invalid psbts. we **MUST** keep the same format for consistency to make life of other teams/sdk devs easier. in general the rule is: if unit tests applies on protocol/psbt validation functions, we follow this JSON format fixtures.

louisinger (changes requested):

louisinger (approved):

PR #61, opened 2026-04-13
Finalize VTXO with configured Arkd when Introspector is the last non-arkd signer

### LINKED ISSUE #58: SubmitTx should finalize the offchain tx with configured arkd

chris-ricketts:
Closes #58

## Summary
- wire trusted `ARKD_URL` through config and service construction for finalizer-path arkd calls
- make finalizer-path `SubmitTx` submit and finalize directly against configured arkd
- finalizer role analysis with accumulator-based owned-input consistency checks
- verify non-arkd checkpoint signatures before arkd submission and merge arkd-returned checkpoints by txid
- add unit coverage for finalizer role analysis, checkpoint signature verification, and finalize retry behavior
- update regtest/docker config to provide the introspector arkd address
- rewrite E2E submit/finalizer tests to use one-call introspector submission with indexer-backed success assertions


internal/application/service.go
```diff
@@ -39,6 +36,12 @@ type SignedBatchFinalization struct {
 	CommitmentTx *psbt.Packet
 }
 
+type FinalizeRetryPolicy struct {
+	MaxRetries           int
+	MinDelayMilliseconds int
+	MaxDelayMilliseconds int
```
louisinger:
```suggestion
	MinDelay time.Duration
	MaxDelay time.Duration
```
beside types, should we add exponential backoff ? also in client-lib we used "jitter" percentage instead of a range. check reconnect logic in client-lib : https://github.com/arkade-os/arkd/blob/0bd4b97d389203c93570e26732a8b352c9338f28/pkg/client-lib/internal/utils/reconnect.go#L11-L21 

if possible, I'd reuse as much as possible the same pattern for consistency (easier to maintain)

internal/config/config.go
```diff
@@ -12,31 +12,42 @@ import (
...
+	Datadir                 = "DATADIR"
+	Port                    = "PORT"
+	NoTLS                   = "NO_TLS"
+	TLSExtraIPs             = "TLS_EXTRA_IPS"
+	TLSExtraDomains         = "TLS_EXTRA_DOMAINS"
+	LogLevel                = "LOG_LEVEL"
+	ArkdURL                 = "ARKD_URL"
+	FinalizeMaxRetries      = "FINALIZE_MAX_RETRIES"
+	FinalizeMinRetryDelayMs = "FINALIZE_MIN_RETRY_DELAY_MS"
+	FinalizeMaxRetryDelayMs = "FINALIZE_MAX_RETRY_DELAY_MS"
```
louisinger: nit: it sounds overkill to make the retry policy configurable ? most of the time we harcode a global with default values. then if u reallly need to, u can modify the global var at runtime (it never happened). 

I'd remove it to simplify the config but it's up to you

internal/application/tx.go
```diff
@@ -43,6 +64,23 @@ func (s *service) SubmitTx(ctx context.Context, tx OffchainTx) (*OffchainTx, err
...
+	defer arkdClient.Close()
+
+	arkdInfo, err := arkdClient.GetInfo(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch arkd info: %w", err)
+	}
+	arkdPubKey, err := arkdInfoSignerPubKey(arkdInfo)
+	if err != nil {
+		return nil, err
+	}
```
louisinger: We shoud create the grpcclient only once to save the gRPC handshake (faster SubmitTx). I think it's ok to initialize it in `New` (it means introspector should start after arkd but it sounds ok now, we could improve later). Another advantage : it "validates" arkdURL on startup. we should also call the GetInfo there and then cache the public key (it is not subject to mutation anyway).

PS: the client is handling reconnection if arkd restart/crash and it is safe from concurrent calls so you don't have to care bout it.

the tradeoff is we need to add a Close method to the service and handle it on process termination

louisinger (changes requested):

internal/config/config.go
chris-ricketts: Addressed in 0d4937c190dfd70692a08a51fff5b05453d4e5c3

internal/application/service.go
chris-ricketts: Addressed in 0d4937c190dfd70692a08a51fff5b05453d4e5c3

internal/application/tx.go
chris-ricketts: Addressed in 9379b6de16bd1cd1686b5a5e9ccf0e77e9e0a7ac

internal/application/tx.go
```diff
@@ -97,8 +113,350 @@ func (s *service) SubmitTx(ctx context.Context, tx OffchainTx) (*OffchainTx, err
...
+	if !isFinalizer {
+		return &OffchainTx{
+			ArkTx:       arkPtx,
+			Checkpoints: signedCheckpointTxs,
+		}, nil
+	}
+
+	// we must verify that we have all the required checkpoint signatures before submitting to arkd
+	// otherwise, finalizing with arkd will fail later
+	if err = verifyNonArkdCheckpointSignatures(signedCheckpointTxs, s.arkdPubKey); err != nil {
```
msinkec, line 134: What about when someone delibaretly submits a fully signed checkpoint transaction but doesn't sign the ark tx itself? It would go into a retry loop. You could potentially abuse this and achieve DOS.

msinkec: Ah, nevermind. Arkd already validates all non-arkd sigs in the SubmitTx call.

internal/config/config.go
```diff
@@ -19,6 +20,7 @@ const (
 	TLSExtraIPs     = "TLS_EXTRA_IPS"
 	TLSExtraDomains = "TLS_EXTRA_DOMAINS"
 	LogLevel        = "LOG_LEVEL"
+	ArkdURL         = "ARKD_URL"
```
louisinger, line 23: let's update README "Configuration" : https://github.com/ArkLabsHQ/introspector?tab=readme-ov-file#configuration

internal/application/tx.go
louisinger, line 134: can we reuse [arkd/pkg/ark-lib/script/VerifyTapscriptSigs](https://github.com/arkade-os/arkd/blob/b5217030543be25fe241e2b70ddaa1d49b05f124/pkg/ark-lib/script/verify.go#L20) ?

internal/application/tx.go
```diff
@@ -97,8 +113,350 @@ func (s *service) SubmitTx(ctx context.Context, tx OffchainTx) (*OffchainTx, err
...
+}
+
+func (a *finalizerAccumulator) isFinalizer() (bool, error) {
+	if len(a.vins) == 0 {
+		return false, nil
+	}
+	referenceVin := a.vins[0]
+	referenceIsLast, ok := a.isLastByVin[referenceVin]
+	if !ok {
+		panic(fmt.Sprintf("missing finalizer state for input %d", referenceVin))
```
louisinger: I know the panics are unreachable but let's return an error for safety ?

internal/application/tx.go
```diff
@@ -97,8 +113,350 @@ func (s *service) SubmitTx(ctx context.Context, tx OffchainTx) (*OffchainTx, err
...
+			lastNonArkdSigner := script.ClosurePubKeys()[nClosurePubKeys-2]
+			lastNonArkdSignerXonly := schnorr.SerializePubKey(lastNonArkdSigner)
+			isLastForEntry = bytes.Equal(lastNonArkdSignerXonly, tweakedSignerPublicKeyXOnly)
+		} else {
+			isLastForEntry = bytes.Equal(lastSignerXOnly, tweakedSignerPublicKeyXOnly)
+		}
+	}
+
+	a.isLastByVin[vin] = isLastForEntry
+}
```
louisinger:
```suggestion
func (a *finalizerAccumulator) checkScript(vin uint16, script *arkade.ArkadeScript) error {
	a.vins = append(a.vins, vin)

	nClosurePubKeys := len(script.ClosurePubKeys())
	tweakedSignerPublicKeyXOnly := schnorr.SerializePubKey(script.PubKey())
	if nClosurePubKeys < 2 {
    // the script comes from ReadArkadeScript so it should be a valid forfeit closure with at least server + tweaked key
    return fmt.Errorf("malformed script %x", script.script)
  }

  lastSigner := script.ClosurePubKeys()[nClosurePubKeys-1]
  lastSignerXOnly := schnorr.SerializePubKey(lastSigner)

  // if arkd is the last signer, check the second-to-last
  if bytes.Equal(lastSignerXOnly, a.arkdPubKeyXonly) {
    lastNonArkdSigner := script.ClosurePubKeys()[nClosurePubKeys-2]
    lastNonArkdSignerXonly := schnorr.SerializePubKey(lastNonArkdSigner)
    a.isLastByVin[vin] = bytes.Equal(lastNonArkdSignerXonly, tweakedSignerPublicKeyXOnly)
    return nil
  }

  a.isLastByVin[vin] = bytes.Equal(lastSignerXOnly, tweakedSignerPublicKeyXOnly)
}
```

2 suggestions :
* nit: remove "else"
* fail fast if nPubKeys < 2 (n=0 is not possible and n=1 would mean arkd pubkey only)

louisinger (changes requested):
let's update "SubmitTx" section in README to document the last signer behavior : https://github.com/ArkLabsHQ/introspector?tab=readme-ov-file#submittx

internal/application/tx.go
```diff
@@ -97,8 +113,350 @@ func (s *service) SubmitTx(ctx context.Context, tx OffchainTx) (*OffchainTx, err
...
+		attempt++
+
+		if err := s.arkdClient.FinalizeTx(ctx, txid, checkpoints); err == nil {
+			return nil
+		} else {
+			log.WithField("txid", txid).WithField("attempt", attempt).Errorf("finalizing tx failed: %s", err)
+		}
+
+		delay := applyJitter(backoffDelay, retryConfig.Jitter)
+		backoffDelay = max(retryConfig.MaxDelay, backoffDelay*time.Duration(retryConfig.Multiplier))
```
msinkec: Shouldn't this be min(...) ?

internal/application/tx.go
chris-ricketts: I don't think so because it [skips inputs without any signatures](https://github.com/arkade-os/arkd/blob/b5217030543be25fe241e2b70ddaa1d49b05f124/pkg/ark-lib/script/verify.go#L148-L151) and the introspector will only add signatures to checkpoints referenced in the packet.

If there is a non-introspector checkpoint that is not signed and the introspector is the finalizer, `arkd.SubmitTx` will succeed but `arkd.FinalizeTx` will fail.

internal/config/config.go
chris-ricketts: Addressed in 3e42460ffe292d1eb72844e07054921eb4fc616a

Same for https://github.com/ArkLabsHQ/introspector/pull/61#pullrequestreview-4111087601

internal/application/tx.go
chris-ricketts: Addressed in 9b4ecec110b3c1c68f8e8462d456c4f09a43fa1f

internal/application/tx.go
chris-ricketts: Addressed in 9b4ecec110b3c1c68f8e8462d456c4f09a43fa1f

internal/application/tx.go
chris-ricketts: Yup 🤦  good spot. Fixed in ed027e50545e6a14d862461c6098073b15548da4

internal/application/tx.go
louisinger: indeed, should have an option to fail instead of skipping.

let's avoid code duplication and use it anyway, we'll update once ark-lib is fixed : https://github.com/arkade-os/arkd/issues/1013

anyway it is "safe" because arkd will fail on SubmitTx

chris-ricketts: `arkd.SubmitTx` fails on missing signatures on the ark tx but doesn't check the checkpoints

louisinger (approved):

PR #65, opened 2026-04-15
SubmitOnchainTx RPC

### LINKED ISSUE #31: Unilateral exit integration test
https://github.com/ArkLabsHQ/introspector/pull/30 includes support of CSV tapscript ark closures, used for unilateral exit scripts. We should add e2e test verifying a VTXO owner is able to unilateral exit with introspector (spending the exit CSV script).

### LINKED ISSUE #59: Onchain execution

louisinger:
Add SubmitOnchainTx RPC allowing to spend arkade script outside of ark offchain env. For safety reason, the introspector will reject any transaction spending a forfeit closure (to avoid bypassing the checks made by SubmitTx).

it closes #59 
it closes #31 

@chris-ricketts @msinkec please review


api-spec/protobuf/introspector/v1/service.proto
chris-ricketts, line 24: Should this be renamed to `SubmitOffchainTx` so it less ambiguous now there is also `SubmitOnchainTx`?

internal/application/prevout.go
```diff
@@ -114,6 +142,40 @@ func decodePrevoutTxsFromPSBT(ptx *psbt.Packet) (map[int]*wire.MsgTx, error) {
 	return prevoutTxs, nil
 }
```
chris-ricketts:
```suggestion
func decodePrevoutTxsFromField(ptx *psbt.Packet, field txutils.ArkPsbtFieldCoder[wire.MsgTx]) (map[int]*wire.MsgTx, error) {
	if len(ptx.Inputs) != len(ptx.UnsignedTx.TxIn) {
		return nil, fmt.Errorf("malformed psbt")
	}

	prevoutTxs := make(map[int]*wire.MsgTx)

	for inputIndex := range ptx.Inputs {
		fields, err := txutils.GetArkPsbtFields(ptx, inputIndex, field)
		if err != nil {
			return nil, fmt.Errorf("failed to decode prevout tx for input %d: %w", inputIndex, err)
		}

		if len(fields) == 0 {
			continue
		}
		if len(fields) > 1 {
			return nil, fmt.Errorf("multiple prevout tx fields found for input %d", inputIndex)
		}

		prevTx := fields[0]
		prevTxCopy := prevTx
		prevoutTxs[inputIndex] = &prevTxCopy
	}

	return prevoutTxs, nil
}

func decodePrevoutTxsFromPSBT(ptx *psbt.Packet) (map[int]*wire.MsgTx, error) {
	return decodePrevoutTxsFromField(ptx, arkade.PrevArkTxField)
}

func decodeOnchainPrevoutTxsFromPSBT(ptx *psbt.Packet) (map[int]*wire.MsgTx, error) {
	return decodePrevoutTxsFromField(ptx, arkade.PrevoutTxField)
}
```

The only difference between the two functions is the field

internal/application/prevout.go
```diff
@@ -114,6 +142,40 @@ func decodePrevoutTxsFromPSBT(ptx *psbt.Packet) (map[int]*wire.MsgTx, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	prevoutTxs, err := decodeOnchainPrevoutTxsFromPSBT(ptx)
+	if err != nil {
+		return nil, err
+	}
+
+	prevOutArkTxs := make(map[wire.OutPoint]*wire.MsgTx, len(prevoutTxs))
```
chris-ricketts: I would maybe rename to something like `prevOutTxMap` or anything without `Ark` in it.

internal/application/prevout.go
louisinger: nice catch, done by 4d6629e689f03d664ebc223171acbf8543a158e8
(I only kept `decodePrevoutTxsFromField` and used it directly in the main functions)

internal/application/prevout.go
louisinger: done af2ca0305aab8ade801922c1525387a2485023d8
(+ remove "fromPSBT" in the factories functions, clear enough without it IMO)

api-spec/protobuf/introspector/v1/service.proto
louisinger: I personally like "SubmitTx" for offchain because it is the same RPC in arkd, but it makes sense. 
anyway, let's keep this in mind, we'll do a review session with Andrew once we're done with 0.0.1 backlog. APIs will need its approval anyway.

PR #66, opened 2026-04-16
Implement E2E contract ID test

msinkec:
Adds an end-to-end test for the contract identity pattern described in #54. The test deploys a main contract with a unique asset-backed contract ID, preserves fixed packet state across spends, and co-spends it with a reader contract that authenticates the main contract via `OP_INSPECTINASSETLOOKUP`.

Shared VTXO/script/extension helpers were moved into `test/utils_test.go` so the existing counter contract test and the new contract ID test can reuse them.


test/contract_id_test.go
```diff
@@ -0,0 +1,433 @@
...
+var fixedStatePayload = uint64LE(0xdeadbeef)
+
+// TestContractIdWithAssetIdentity exercises the contract identity pattern:
+//   - A main contract carries a unique asset (contract ID) and fixed state.
+//     It enforces output continuation (same script, asset forwarded, state preserved).
+//   - A reader contract is co-spent with the main contract, verifies the main
+//     contract's asset identity via OP_INSPECTINASSETLOOKUP, and reads the state
+//     from the current transaction's packet.
+func TestContractIdWithAssetIdentity(t *testing.T) {
+	ctx := context.Background()
```
louisinger:
```suggestion
	ctx := t.Context()
```

test/contract_id_test.go
```diff
@@ -0,0 +1,428 @@
...
+// Fixed state payload carried by the main contract across spends.
+var fixedStatePayload = uint64LE(0xdeadbeef)
+
+// TestContractIdWithAssetIdentity exercises the contract identity pattern:
+//   - A main contract carries a unique asset (contract ID) and fixed state.
+//     It enforces output continuation (same script, asset forwarded, state preserved).
+//   - A reader contract is co-spent with the main contract, verifies the main
+//     contract's asset identity via OP_INSPECTINASSETLOOKUP, and reads the state
+//     from the current transaction's packet.
+func TestContractIdWithAssetIdentity(t *testing.T) {
```
louisinger, line 37: LGTM 

let's also add a test case where the main contract is spent in settlement. the execution should succeed but the state should be lost because arkd is missing this feature : https://github.com/arkade-os/arkd/issues/1017 

having the test will be use as "reference" use case

msinkec: > LGTM
> 
> let's also add a test case where the main contract is spent in settlement. the execution should succeed but the state should be lost because arkd is missing this feature : [arkade-os/arkd#1017](https://github.com/arkade-os/arkd/issues/1017)
> 
> having the test will be use as "reference" use case

I added a test placeholder to test batch-continuations of recursive covenants. I've chosen the counter contract to be used in this test, since it's the simplest recursive contract with dynamic state. [64dd6d1](https://github.com/ArkLabsHQ/introspector/pull/66/commits/64dd6d16413deaa62adc837e570a86b398cfe465)

I've also removed the extra deployment script for the counter contract since I've deemed it unnecessary. I've also refactored the related tests a bit to re-use helper funcs. [0ab827d](https://github.com/ArkLabsHQ/introspector/pull/66/commits/0ab827dee982f922e9b73cd4ddae262a78c05063)

louisinger (approved):
ACK, waiting for https://github.com/arkade-os/arkd/pull/1022 to merge

PR #69, opened 2026-04-22
Unified BigInt Arithmetic

### LINKED ISSUE #60: Unified big number arithmetic
## 1. Motivation

The introspector currently has three incompatible numeric representations:

- **scriptNum**: int64 internally, 4-byte input limit, sign-magnitude little-endian encoding. Used by standard arithmetic opcodes (OP_ADD, OP_MUL, etc.).
- **LE64**: fixed 8-byte little-endian, two's complement. Used by a parallel set of 64-bit arithmetic opcodes (OP_ADD64, OP_MUL64, etc.) and pushed by introspection opcodes.
- **big.Int**: used ad-hoc by shift opcodes only.

This creates several problems:

1. **Opcode duplication.** Every arithmetic operation exists twice (OP_ADD and OP_ADD64), with different encodings, different error semantics, and different stack transforms. Adding a third operand size (e.g., 256-bit for field arithmetic) would require yet another set of opcodes and conversion opcodes.
2. **Conversion overhead.** Moving values between representations requires explicit conversion opcodes (OP_SCRIPTNUMTOLE64, OP_LE64TOSCRIPTNUM, OP_LE32TOLE64). Introspection results are pushed as LE64 and must be converted before they can participate in scriptNum arithmetic or vice versa.
3. **Missing byte-string bridging.** There is no way to convert between numeric values and fixed-width byte strings. This makes bitwise opcodes (OP_XOR, OP_AND, OP_OR) hard to use with numeric data, since they require equal-length operands and produce raw byte results with no path back to a normalized numeric value.
4. **Limited arithmetic range.** The 4-byte scriptNum limit restricts arithmetic to [-2^31 + 1, 2^31 - 1]. Even LE64 arithmetic caps at ~2^63, which is insufficient for cryptographic field operations or precision-preserving multi-value products.
5. **Special-case parsing paths.** `OP_CHECKLOCKTIMEVERIFY` and `OP_CHECKSEQUENCEVERIFY` currently bypass normal numeric parsing to handle 32-bit locktime/sequence values that exceed the legacy 4-byte scriptNum limit. A unified BigNum model removes this exception.

This spec replaces all three representations with a single unified BigNum type, raises the arithmetic limit to the existing maximum stack element size (520 bytes), and adds OP_NUM2BIN / OP_BIN2NUM to bridge between numeric and byte-string domains.

### Encoding

All numeric values on the stack use the existing sign-magnitude little-endian encoding (the same format scriptNum uses today):

- **Little-endian byte order**, magnitude in the lower bytes
- **Sign bit** in bit 7 of the most significant byte
- **Zero** is encoded as the empty byte array (0 bytes)
- **Minimal encoding required**: no unnecessary leading zero-bytes except when needed for sign extension. Negative zero (`[0x80]`) is not a valid encoding. The existing `checkMinimalDataEncoding` validation applies unchanged.

Encoding examples:

| Value | Bytes          |
| ----- | -------------- |
| 0     | `[]` (empty)   |
| 1     | `[0x01]`       |
| -1    | `[0x81]`       |
| 127   | `[0x7f]`       |
| 128   | `[0x80, 0x00]` |
| -128  | `[0x80, 0x80]` |
| 255   | `[0xff, 0x00]` |
| -255  | `[0xff, 0x80]` |

### Limits

- **Maximum encoded size** of any numeric operand or result: **520 bytes** (equal to the existing `MaxScriptElementSize`).
- **Effective numeric range**: [-(2^4159 - 1), 2^4159 - 1].
- If an arithmetic operation produces a result whose minimal encoding exceeds 520 bytes, the script **fails immediately** (hard error).

### Why Hard-Fail on Overflow

With 520 bytes, the numeric range spans over 1,250 decimal digits. No legitimate contract computation should accidentally exceed this range. If it does, it indicates a bug in the contract logic. Hard-fail makes this class of error impossible to miss, unlike soft-fail models where a forgotten overflow check can cause the script to silently continue on stale operand data.

Additionally, the full range is usable for reducing script size in field arithmetic: intermediate results can accumulate across multiple operations with a single modular reduction at the end, rather than reducing after every step. The 520-byte ceiling provides ample headroom for this pattern. Contracts that need bounded results should use OP_MOD explicitly.

All numeric opcodes, including shift operations, interpret stack bytes using this encoding before execution and re-encode results back into this same minimally-encoded form.

### 3.1 Removed Opcodes (12)

These opcodes are removed entirely. Their byte values become invalid (`OP_UNKNOWN` / `opcodeInvalid`).

**LE64 arithmetic (replaced by unified opcodes):**

| Opcode                  | Hex  | Replacement           |
| ----------------------- | ---- | --------------------- |
| OP_ADD64                | 0xd7 | OP_ADD                |
| OP_SUB64                | 0xd8 | OP_SUB                |
| OP_MUL64                | 0xd9 | OP_MUL                |
| OP_DIV64                | 0xda | OP_DIV                |
| OP_NEG64                | 0xdb | OP_NEGATE             |
| OP_LESSTHAN64           | 0xdc | OP_LESSTHAN           |
| OP_LESSTHANOREQUAL64    | 0xdd | OP_LESSTHANOREQUAL    |
| OP_GREATERTHAN64        | 0xde | OP_GREATERTHAN        |
| OP_GREATERTHANOREQUAL64 | 0xdf | OP_GREATERTHANOREQUAL |

**Conversion opcodes (no longer needed with a single numeric type):**

| Opcode             | Hex  |
| ------------------ | ---- |
| OP_SCRIPTNUMTOLE64 | 0xe0 |
| OP_LE64TOSCRIPTNUM | 0xe1 |
| OP_LE32TOLE64      | 0xe2 |

### 3.2 Modified Opcodes -- Arithmetic

All arithmetic and comparison opcodes change from a 4-byte scriptNum input limit to a 520-byte BigNum input limit. Stack transforms are unchanged. All inputs must be minimally encoded. All results are pushed as minimally-encoded BigNum. If a result exceeds 520 bytes, the script fails.

**Unary operators:**

| Opcode       | Hex  | Stack Transform         |
| ------------ | ---- | ----------------------- |
| OP_1ADD      | 0x8b | `[x] -> [x+1]`          |
| OP_1SUB      | 0x8c | `[x] -> [x-1]`          |
| OP_NEGATE    | 0x8f | `[x] -> [-x]`           |
| OP_ABS       | 0x90 | `[x] -> [abs(x)]`       |
| OP_NOT       | 0x91 | `[x] -> [x==0 ? 1 : 0]` |
| OP_0NOTEQUAL | 0x92 | `[x] -> [x!=0 ? 1 : 0]` |

**Binary arithmetic operators:**

| Opcode | Hex  | Stack Transform      | Notes                                                                         |
| ------ | ---- | -------------------- | ----------------------------------------------------------------------------- |
| OP_ADD | 0x93 | `[x1 x2] -> [x1+x2]` |                                                                               |
| OP_SUB | 0x94 | `[x1 x2] -> [x1-x2]` |                                                                               |
| OP_MUL | 0x95 | `[x1 x2] -> [x1*x2]` |                                                                               |
| OP_DIV | 0x96 | `[x1 x2] -> [x1/x2]` | Truncated division. Quotient only. Division by zero fails the script.         |
| OP_MOD | 0x97 | `[x1 x2] -> [x1%x2]` | Truncated remainder (sign follows dividend). Modulo by zero fails the script. |

### Shift operators

| Opcode    | Hex  | Stack Transform     | Notes |
| --------- | ---- | ------------------- | ----- |
| OP_LSHIFT | 0x98 | `[x n] -> [x << n]` | Both operands BigNum. `n` must be non-negative. `x` is decoded from stack bytes using BigNum sign-magnitude little-endian semantics, shifted numerically, then re-encoded as minimally-encoded sign-magnitude little-endian. Result must fit in 520 bytes. Implementations may early-fail when `n > 4160` and `x != 0`, since the result is guaranteed to exceed 520 bytes. |
| OP_RSHIFT | 0x99 | `[x n] -> [x >> n]` | Both operands BigNum. `n` must be non-negative. `x` is decoded from stack bytes using BigNum sign-magnitude little-endian semantics, shifted numerically, then re-encoded as minimally-encoded sign-magnitude little-endian. Arithmetic shift: rounds toward negative infinity (e.g., `-1 >> 1 == -1`, `-7 >> 1 == -4`). Shifting by more than the value's bit-width produces `0` for non-negative values and `-1` for negative values. |

**Compatibility note:** This is a behavior change relative to the current implementation. Existing shift opcodes decode stack bytes incorrectly by interpreting them as unsigned big-endian values via `big.Int.SetBytes`, then re-encoding via `big.Int.Bytes()`. Under this spec, shifts operate on properly decoded BigNum values using sign-magnitude little-endian stack semantics. Scripts that previously used shifts on anything other than small positive single-byte values may therefore produce different results. This change is intentional and treated as a bug fix.

#### Shift examples

| Stack item    | BigNum value | Op     | Correct result |
| ------------- | ------------ | ------ | -------------- |
| `[0x05]`      | `5`          | `<< 1` | `[0x0a]`       |
| `[0xff,0x00]` | `255`        | `<< 1` | `[0xfe,0x01]`  |
| `[0xff]`      | `-127`       | `<< 1` | `[0xfe,0x80]`  |
| `[0x04,0x80]` | `-4`         | `>> 1` | `[0x82]`       |
| `[0x81]`      | `-1`         | `>> 1` | `[0x81]`       |
| `[0x07,0x80]` | `-7`         | `>> 1` | `[0x84]`       |

**Comparison and boolean operators:**

| Opcode                | Hex  | Stack Transform                       |
| --------------------- | ---- | ------------------------------------- |
| OP_BOOLAND            | 0x9a | `[x1 x2] -> [x1!=0 && x2!=0 ? 1 : 0]` |
| OP_BOOLOR             | 0x9b | `[x1 x2] -> [x1!=0 || x2!=0 ? 1 : 0]` |
| OP_NUMEQUAL           | 0x9c | `[x1 x2] -> [x1==x2 ? 1 : 0]`         |
| OP_NUMEQUALVERIFY     | 0x9d | `[x1 x2] -> []` (fails if x1 != x2)   |
| OP_NUMNOTEQUAL        | 0x9e | `[x1 x2] -> [x1!=x2 ? 1 : 0]`         |
| OP_LESSTHAN           | 0x9f | `[x1 x2] -> [x1<x2 ? 1 : 0]`          |
| OP_GREATERTHAN        | 0xa0 | `[x1 x2] -> [x1>x2 ? 1 : 0]`          |
| OP_LESSTHANOREQUAL    | 0xa1 | `[x1 x2] -> [x1<=x2 ? 1 : 0]`         |
| OP_GREATERTHANOREQUAL | 0xa2 | `[x1 x2] -> [x1>=x2 ? 1 : 0]`         |
| OP_MIN                | 0xa3 | `[x1 x2] -> [min(x1,x2)]`             |
| OP_MAX                | 0xa4 | `[x1 x2] -> [max(x1,x2)]`             |
| OP_WITHIN             | 0xa5 | `[x min max] -> [min<=x<max ? 1 : 0]` |

### 3.3 Modified Opcodes -- Locktime / Sequence Verification

`OP_CHECKLOCKTIMEVERIFY` and `OP_CHECKSEQUENCEVERIFY` (including any aliases/macros referred to as `OP_CLTV` / `OP_CSV`) now parse their stack operand as a **BigNum**, not as a legacy 4-byte scriptNum and not via the old 5-byte special case.

- **Input encoding:** minimally-encoded sign-magnitude little-endian BigNum
- **Maximum encoded size:** 520 bytes
- **Negative operands:** invalid, as before
- **Consensus semantics:** unchanged

The transaction fields being checked remain protocol-defined 32-bit values (`nLockTime` and `nSequence`). Implementations compare the BigNum operand against the corresponding transaction field after widening that field to BigNum form.

This removes the historical exception where these opcodes bypassed normal integer parsing in order to accept 5-byte operands for values above the legacy 4-byte scriptNum limit.

Operands larger than the relevant protocol field range are still valid BigNum encodings, but they will fail the opcode's normal verification logic when compared against the transaction's 32-bit field.

### 3.4 Modified Opcodes -- Introspection

These opcodes currently push values as 8-byte LE64. They change to push minimally-encoded BigNum values instead. Since satoshi values and asset amounts are unsigned 64-bit integers, their BigNum encoding will be at most 9 bytes. These values feed directly into arithmetic opcodes without any conversion.

| Opcode                   | Hex  | Description                   |
| ------------------------ | ---- | ----------------------------- |
| OP_INSPECTINPUTVALUE     | 0xc9 | Input satoshi value           |
| OP_INSPECTOUTPUTVALUE    | 0xcf | Output satoshi value          |
| OP_INSPECTASSETGROUP     | 0xeb | Asset group amounts           |
| OP_INSPECTASSETGROUPSUM  | 0xec | Asset group sum amounts       |
| OP_INSPECTOUTASSETAT     | 0xee | Output asset amount at index  |
| OP_INSPECTOUTASSETLOOKUP | 0xef | Output asset amount by lookup |
| OP_INSPECTINASSETAT      | 0xf1 | Input asset amount at index   |
| OP_INSPECTINASSETLOOKUP  | 0xf2 | Input asset amount by lookup  |

#### OP_NUM2BIN

**Stack transform:** `[num size] -> [bytes]`

Converts a BigNum to a byte string of exactly `size` bytes.

1. Pop `size` (small integer, 4-byte limit, must be non-negative and <= 520).
2. Pop `num` (BigNum).
3. Encode `num` as sign-magnitude little-endian, zero-padded to exactly `size` bytes (padding bytes are inserted between the magnitude and the sign-extension byte).
4. If `num` cannot fit in `size` bytes, the script fails.

Examples:

| Input              | Output                           |
| ------------------ | -------------------------------- |
| `5 4 OP_NUM2BIN`   | `[0x05, 0x00, 0x00, 0x00]`       |
| `-5 4 OP_NUM2BIN`  | `[0x05, 0x00, 0x00, 0x80]`       |
| `0 4 OP_NUM2BIN`   | `[0x00, 0x00, 0x00, 0x00]`       |
| `255 1 OP_NUM2BIN` | script fails (255 needs 2 bytes) |

#### OP_BIN2NUM

**Stack transform:** `[bytes] -> [num]`

Converts an arbitrary byte string to a minimally-encoded BigNum. Operates directly on the byte vector without decoding into a numeric type and re-encoding.

1. Pop `bytes` (byte array, max 520 bytes).
2. Minimally encode the byte vector in place: strip trailing zero-bytes (working backward from the most significant byte), preserving the sign bit. Normalize negative zero (`[0x80]`) to positive zero (empty byte array).
3. Push the minimally-encoded result.

The operation never constructs an intermediate numeric value. It is purely a byte-level normalization: strip padding, fix the sign byte, push.

**OP_BIN2NUM is the only opcode that accepts non-minimally-encoded numeric input.** This is its purpose: it is the "normalize this byte string into a valid number" conversion. All other numeric opcodes require minimal encoding on input.

Examples:

| Input                                 | Output                       |
| ------------------------------------- | ---------------------------- |
| `[0x05, 0x00, 0x00, 0x00] OP_BIN2NUM` | `[0x05]`                     |
| `[0x00, 0x00, 0x00, 0x80] OP_BIN2NUM` | `[]` (negative zero -> zero) |
| `[0x05, 0x00, 0x00, 0x80] OP_BIN2NUM` | `[0x85]`                     |

### 3.6 Opcodes That Retain Small-Integer Parsing

The following opcodes continue to parse their numeric arguments with the legacy 4-byte limit. These are indices, counts, sizes, and protocol-specific fields where large values are meaningless or undesirable:

- Stack manipulation: `OP_PICK`, `OP_ROLL`
- String operations: `OP_SPLIT` (index), `OP_NUM2BIN` (size argument)
- Flow control: `OP_IF`, `OP_NOTIF` (interpreted as boolean)
- Multisig: key count and signature count parameters
- Introspection indices: input/output index parameters

`OP_CHECKLOCKTIMEVERIFY` and `OP_CHECKSEQUENCEVERIFY` are **not** included in this list. They now consume BigNum operands like other numeric opcodes.

## 4. Performance Analysis

All benchmarks measured in Go on Apple M4 Pro using `math/big` and `crypto/ecdsa`.

### Measured Results

| Operation                 | Time      | Heap Allocs | Ratio vs Sig Verify |
| ------------------------- | --------- | ----------- | ------------------- |
| Native int64 multiply     | 0.23 ns   | 0           | ~144,000x cheaper   |
| `math/big` int64 multiply | 5.8 ns    | 0           | ~5,700x cheaper     |
| 520x520-byte ADD          | 15 ns     | 0           | ~2,200x cheaper     |
| 520x520-byte MUL          | 1,080 ns  | 0           | ~31x cheaper        |
| 520x520-byte DIV          | 2,580 ns  | 1           | ~13x cheaper        |
| 520x520-byte MOD          | 2,550 ns  | 1           | ~13x cheaper        |
| Double-SHA256 (200 bytes) | 116 ns    | 0           | ~287x cheaper       |
| ECDSA P256 Verify         | 33,300 ns | 19          | baseline            |

Note: ECDSA uses Go's P256, not secp256k1. The actual OP_CHECKSIG in the introspector uses btcec/secp256k1 and also computes sighash preimages, making it even more expensive than this baseline.

### Key Takeaways

1. The absolute worst case -- a 520x520-byte multiplication -- is **31x cheaper** than a single signature verification. Division is **13x cheaper**.
2. The int64 fast path matters: native int64 multiply is **25x faster** than routing through `math/big` for the same small values. The common case (satoshi arithmetic, counters) stays near-zero cost.
3. ADD/SUB at 520 bytes is essentially free (~15 ns, zero heap allocations).
4. Uniform per-opcode pricing is reasonable at this limit. The worst-case MUL is bounded at ~1 us -- a script would need >30 worst-case multiplications to equal one signature check.

### 5.1 Constant-Product AMM (x * y = k)

A constant-product automated market maker enforces the invariant `x * y = k` across a swap, where `x` and `y` are token reserves. Precision is critical -- any rounding error in the invariant check can be exploited to drain the pool over time.

**Problem without BigNum:** If reserves are 64-bit satoshi values, their product can exceed 64 bits. A pool with two reserves of ~10^18 sats each overflows on multiplication. The contract either needs to work with reduced precision (lossy) or split the computation across multiple opcodes with carry logic (complex and error-prone).

**With BigNum:** The product of two 8-byte values fits in at most 16 bytes -- trivially within the 520-byte limit. The invariant check becomes straightforward:

```
// Stack: [x_old, y_old, x_new, y_new]
// Verify: x_old * y_old <= x_new * y_new

OP_2OVER                      // copy x_old, y_old to top
OP_MUL                        // k_old = x_old * y_old  (up to 16 bytes, no overflow)
OP_SWAP
OP_ROT
OP_MUL                        // k_new = x_new * y_new
OP_LESSTHANOREQUAL            // k_old <= k_new
OP_VERIFY
```

No precision loss, no carry logic, no conversion opcodes.

### 5.2 On-Chain Elliptic Curve Field Arithmetic

Verifying or computing elliptic curve operations in script requires modular arithmetic over a field prime. For secp256k1 this is `p = 2^256 - 2^32 - 977`, a 32-byte number. Field operations include:

- **Addition:** `(a + b) mod p`
- **Multiplication:** `(a * b) mod p` -- the product of two 32-byte values is up to 64 bytes
- **Inversion:** via Fermat's little theorem, `a^(p-2) mod p`, built from repeated squaring and multiplication

**Problem without BigNum:** Impossible with 4-byte scriptNum. With LE64 you cannot even represent a 32-byte field element.

**With BigNum:** A field multiply is two opcodes:

```
// Stack: [a, b, p]
OP_ROT OP_ROT                 // [p, a, b]
OP_MUL                        // [p, a*b]  (up to 64 bytes, within 520-byte limit)
OP_SWAP
OP_MOD                        // [(a*b) mod p]
```

The full 520-byte range means intermediate products from multiple field operations can accumulate before reducing, since the intermediates fit comfortably. This enables:

- On-chain signature recovery and alternative signing scheme verification
- Pedersen commitment arithmetic (proving relationships between committed values)
- Custom hash-to-curve implementations

### 6.1 The BigNum Type

A tagged union holding either a native `int64` or a `*big.Int`:

```go
type BigNum struct {
    small  int64
    big    *big.Int
    useBig bool
}
```

Invariant: when `useBig` is false, the value lives in `small`. When true, it lives in `big`. Operations check both operands and take the fastest applicable path.

### 6.2 Decoding from Stack Bytes

A new `MakeBigNum` function handles all BigNum-consuming opcode inputs, including arithmetic, comparisons, shifts, and `OP_CHECKLOCKTIMEVERIFY` / `OP_CHECKSEQUENCEVERIFY`. The existing `MakeScriptNum` remains only for opcodes that intentionally preserve legacy 4-byte small-integer parsing (`OP_PICK`, `OP_ROLL`, `OP_SPLIT`, `OP_NUM2BIN` size, multisig counts, introspection indices, etc.).

```go
MakeBigNum(v []byte, requireMinimal bool, maxLen int) (BigNum, error)
```

Decision logic:

1. `len(v) > maxLen` -> error (`maxLen` = 520 for arithmetic opcodes).
2. If `requireMinimal`: run existing `checkMinimalDataEncoding(v)`.
3. `len(v) == 0` -> return zero BigNum (int64 path).
4. `len(v) <= 8` -> decode into `int64` using the existing sign-magnitude LE decoding.
5. `len(v) > 8` -> decode into `*big.Int`: interpret bytes as sign-magnitude LE, extract sign bit from MSB, construct magnitude, negate if needed.

The 8-byte boundary is clean: all minimally-encoded sign-magnitude values up to 8 bytes fit in `int64` (max +/-(2^63 - 1)). At 9+ bytes the minimum magnitude is 2^63, which requires `big.Int`.

### 6.3 Encoding to Stack Bytes

```go
func (n BigNum) Bytes() []byte
```

- If int64 path: use the existing scriptNum encoding logic (int64 -> sign-magnitude LE).
- If big.Int path: extract sign, convert magnitude to little-endian bytes, apply sign bit to MSB, add sign-extension byte if MSB's high bit conflicts with sign.

After encoding, check `len(result) > 520` -> return script error.

### 6.4 Arithmetic Fast Paths

For each binary operation, the dispatch follows this pattern:

```
if both operands are int64:
    attempt native int64 operation with overflow detection
    if no overflow: return result as int64 BigNum
    else: promote both to big.Int and fall through
big.Int path:
    perform operation via math/big
    return result as big.Int BigNum
```

Overflow detection for the int64 path:

| Operation | Overflow condition                |
| --------- | --------------------------------- |
| Add       | `(result ^ a) & (result ^ b) < 0` |
| Sub       | `(a ^ b) & (a ^ result) < 0`      |
| Mul       | `a != 0 && result/a != b`         |
| Div       | `a == MinInt64 && b == -1`        |
| Mod       | `a == MinInt64 && b == -1`        |
| Negate    | `a == MinInt64`                   |

When overflow is detected, promote both operands to `big.Int` and redo the operation. This is rare (only at the int64 boundary) and the cost of the wasted attempt is negligible.

**Promotion is one-way.** Once a BigNum value is on the `big.Int` path, it never demotes back to `int64`, even if the result would fit. This avoids redundant range checks on every operation and keeps the logic simple. A value starts on whichever path its decoded byte length dictates (≤8 bytes -> int64, >8 bytes -> big.Int) and can only move upward.

### 6.5 Shift Opcode Implementation

Shift opcodes must operate on decoded BigNum values, not on raw byte vectors interpreted through `big.Int.SetBytes`.

The current implementation pattern:

```go
value := new(big.Int).SetBytes(x)
result := value.Lsh(value, uint(n)) // or Rsh
vm.dstack.PushByteArray(result.Bytes())
```

is incorrect for stack numerics because `SetBytes` and `Bytes` use **unsigned big-endian** encoding, while stack numerics use **minimally-encoded sign-magnitude little-endian**.

Correct implementation is:

1. Pop `n` as a BigNum per opcode semantics, then verify `n >= 0`.
2. Pop `x` via `PopBigNum(520)` so the value is decoded from stack bytes using sign-magnitude little-endian rules.
3. Perform the shift on the numeric value.
4. Re-encode via `PushBigNum`, which produces minimally-encoded sign-magnitude little-endian output.
5. For `OP_LSHIFT`, fail if the encoded result exceeds 520 bytes.
6. For `OP_RSHIFT`, use **arithmetic right shift** semantics: division by `2^n` rounded toward negative infinity.

For left shift, the operation is equivalent to multiplication by `2^n`:

```go
result = x * 2^n
```

For right shift, the operation is equivalent to floor division by `2^n`:

```go
result = floor(x / 2^n)
```

This matters for negative values:

- `-1 >> 1 == -1`
- `-7 >> 1 == -4`

Shifting right by more than the operand's bit-width yields:

- `0` for non-negative values
- `-1` for negative values

This follows directly from arithmetic right shift semantics.

### 6.6 Stack Interface

Add parallel methods alongside the existing `PopInt` / `PushInt`:

```go
PopBigNum(maxLen int) (BigNum, error)  // decode top of stack as BigNum
PeekBigNum(maxLen int) (BigNum, error) // decode top of stack as BigNum without popping
PushBigNum(n BigNum) error             // encode and push; error if > 520 bytes
```

`PopBigNum` and `PeekBigNum` call `MakeBigNum(v, true, maxLen)` with minimal encoding enforced. This is the required input path for arithmetic, comparison, and shift opcodes, as well as `OP_CHECKLOCKTIMEVERIFY` / `OP_CHECKSEQUENCEVERIFY`. `PushBigNum` calls `n.Bytes()`, checks the 520-byte limit, and pushes.

The existing `PopInt` / `PushInt` remain for opcodes that still need small-integer semantics.

### 6.7 Introspection Opcode Changes

Introspection opcodes change from pushing raw LE64 bytes to pushing via `PushBigNum`:

```go
// Old: pushAmountLE64(vm, amount)
// New: vm.dstack.PushBigNum(BigNumFromUint64(amount))
```

`BigNumFromUint64` constructs a BigNum on the int64 fast path. Satoshi values and asset amounts always fit in uint64; positive values up to `2^63 - 1` use the int64 path, larger values use big.Int (though satoshi values will not reach this in practice).

### 6.8 `OP_CHECKLOCKTIMEVERIFY` / `OP_CHECKSEQUENCEVERIFY` Changes

These opcodes no longer use a special 5-byte numeric decoding path.

Old behavior relied on peeking the raw byte array and decoding it outside the normal integer path so that `nLockTime` / `nSequence` values above the legacy 4-byte scriptNum ceiling could still be represented.

New behavior is:

```go
arg, err := vm.dstack.PeekBigNum(520)
if err != nil {
    return err
}
```

The transaction field being checked is widened to BigNum for comparison:

```go
txLockTime := BigNumFromUint64(uint64(vm.tx.LockTime))
txSequence := BigNumFromUint64(uint64(txIn.Sequence))
```

All existing consensus checks remain unchanged; only operand parsing changes.

Obsolete comment to remove:

```go
// The current transaction locktime is a uint32 resulting in a maximum
// locktime of 2^32-1 (the year 2106).  However, scriptNums are signed
// and therefore a standard 4-byte scriptNum would only support up to a
// maximum of 2^31-1 (the year 2038).  Thus, a 5-byte scriptNum is used
// here since it will support up to 2^39-1 which allows dates beyond the
// current locktime limit.
//
// PeekByteArray is used here instead of PeekInt because we do not want
// to be limited to a 4-byte integer for reasons specified above.
```

Replacement:

```go
// OP_CHECKLOCKTIMEVERIFY / OP_CHECKSEQUENCEVERIFY now parse their operand
// as a minimally-encoded BigNum, so the old 5-byte scriptNum exception is
// no longer needed.
```

### 6.9 OP_NUM2BIN Implementation

```
Pop size (small integer, 4-byte limit, must be non-negative and <= 520)
Pop num  (BigNum)
Encode num as sign-magnitude LE
Pad to exactly size bytes (insert zero-bytes between magnitude and sign byte)
If num cannot fit in size bytes: script error
Push result as raw byte array
```

### 6.10 OP_BIN2NUM Implementation

```
Pop bytes (raw byte array via PopByteArray, NOT PopBigNum)
Call MinimallyEncode(bytes)   // byte-level normalization, no numeric decode
Push result as raw byte array
```

OP_BIN2NUM operates directly on the byte vector. It does not decode into a BigNum and re-encode. The `MinimallyEncode` function strips trailing zero-bytes (preserving the sign bit) and normalizes negative zero to empty. This matches the existing `checkMinimalDataEncoding` validation in reverse: where `checkMinimalDataEncoding` rejects non-minimal input, `MinimallyEncode` produces minimal output from arbitrary input.

msinkec:
This PR unifies Arkade VM numeric handling around `BigNum` as described in https://github.com/ArkLabsHQ/introspector/issues/60:

  - Adds the `BigNum` type with int64 fast path, big.Int fallback, minimal sign-magnitude encoding,
  arithmetic, shifts, and stack helpers.
  - Migrates arithmetic, boolean, comparison, min/max/within, CLTV/CSV, and asset/value introspection paths to
  use BigNum.
  - Updates introspection amount outputs to push minimally encoded BigNum values directly.
  - Deletes legacy LE64 arithmetic/conversion opcodes and updates scripts/tests to use unified BigNum
  arithmetic.
  - Adds `OP_NUM2BIN` (`0xd7`) and `OP_BIN2NUM` (`0xd8`) for fixed-width byte-string bridging.
  - Updates README opcode docs and test coverage for the new numeric behavior.

Closes https://github.com/ArkLabsHQ/introspector/issues/60


pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+	return BigNum{small: v}
+}
+
+// BigNumFromUint64 constructs a BigNum from an unsigned 64-bit value. Values
+// up to math.MaxInt64 use the int64 fast path; larger values promote to big.
+func BigNumFromUint64(v uint64) BigNum {
+	const int64Max = uint64(1<<63 - 1)
+	if v <= int64Max {
+		return BigNum{small: int64(v)}
+	}
```
chris-ricketts: Nit: we can use `math.MaxInt64` directly for easier readability

```suggestion
	if v <= math.MaxInt64 {
		return BigNum{small: int64(v)}
	}
```

pkg/arkade/bignum_test.go
```diff
@@ -0,0 +1,396 @@
...
+	}
+	// Value at uint64 max (> int64 max) must use big path.
+	n = BigNumFromUint64(^uint64(0))
+	if !n.useBig {
+		t.Fatalf("max uint64 must use big path")
+	}
+	want := new(big.Int).SetUint64(^uint64(0))
+	if n.big.Cmp(want) != 0 {
+		t.Fatalf("big = %s, want %s", n.big, want)
+	}
```
chris-ricketts: Nit: use constants for better readability

Also: add some boundary test cases:

```suggestion
	n = BigNumFromUint64(math.MaxInt64)
	if n.useBig {
		t.Fatalf("max int64 should be on int64 path")
	}
	if n.small != math.MaxInt64 {
		t.Fatalf("small = %d, want %d", n.small, math.MaxInt64)
	}
	// Value at uint64 max (> int64 max) must use big path.
	n = BigNumFromUint64(math.MaxInt64 + 1)
	if !n.useBig {
		t.Fatalf("max int64 must use big path")
	}
	want := new(big.Int).SetUint64(math.MaxInt64 + 1)
	if n.big.Cmp(want) != 0 {
		t.Fatalf("big = %s, want %s", n.big, want)
	}
	n = BigNumFromUint64(math.MaxUint64)
	if !n.useBig {
		t.Fatalf("max uint64 must use big path")
	}
	want = new(big.Int).SetUint64(math.MaxUint64)
	if n.big.Cmp(want) != 0 {
		t.Fatalf("big = %s, want %s", n.big, want)
	}
```

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+		return n.big
+	}
+	return big.NewInt(n.small)
+}
+
+// Add returns n + m. Promotes to big on int64 overflow.
+func (n BigNum) Add(m BigNum) BigNum {
+	if !n.useBig && !m.useBig {
+		r := n.small + m.small
+		// Overflow when sign of both operands matches and differs from result.
```
chris-ricketts: Maybe a comment for the people with rusty bit twiddling skills:

```suggestion
// Detect signed overflow for n.small + m.small.
//
// In two’s complement, overflow occurs iff both operands have the same sign
// and the result has the opposite sign.
//
// (r ^ n.small) has sign bit set if r and n differ in sign.
// (r ^ m.small) has sign bit set if r and m differ in sign.
// If both are set, the AND keeps the sign bit → negative → overflow.
//
// Example (64-bit):
//   n = 01111111...11111111  (MaxInt64)
//   m = 00000000...00000001
//   r = 10000000...00000000  (MinInt64, overflow)
//
//   r ^ n = 11111111...11111111  (sign bit = 1)
//   r ^ m = 10000000...00000001  (sign bit = 1)
//   AND   = 10000000...00000001  (negative → overflow)
//
// So:
//   ((r ^ n.small) & (r ^ m.small)) < 0   => overflow
//   ((r ^ n.small) & (r ^ m.small)) >= 0  => no overflow
```

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+		}
+	}
+	return BigNum{big: new(big.Int).Add(n.asBig(), m.asBig()), useBig: true}
+}
+
+// Sub returns n - m. Promotes to big on int64 overflow.
+func (n BigNum) Sub(m BigNum) BigNum {
+	if !n.useBig && !m.useBig {
+		r := n.small - m.small
+		if (n.small^m.small)&(n.small^r) >= 0 {
```
chris-ricketts, line 166: Another comment for the left-curvers (like me):
```suggestion
// Detect signed overflow for n.small - m.small.
//
// In two's complement, subtraction overflows iff:
//   - n and m have different signs, and
//   - the result has a different sign from n.
//
// (n.small ^ m.small) has sign bit set if n and m differ in sign.
// (n.small ^ r) has sign bit set if r and n differ in sign.
// If both are set, the AND keeps the sign bit -> negative -> overflow.
//
// Example (64-bit):
//   n = 01111111...11111111  (MaxInt64)
//   m = 11111111...11111111  (-1)
//   r = 10000000...00000000  (MinInt64, overflow)
//
//   n ^ m = 10000000...00000000
//   n ^ r = 11111111...11111111
//   AND   = 10000000...00000000  (negative -> overflow)
if (n.small^m.small)&(n.small^r) >= 0 {
```

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+// Mod returns truncated n % m (sign follows dividend). Caller MUST verify
+// m is non-zero first.
+func (n BigNum) Mod(m BigNum) BigNum {
+	if !n.useBig && !m.useBig {
+		if !(n.small == -9223372036854775808 && m.small == -1) {
+			return BigNum{small: n.small % m.small}
+		}
+	}
+	return BigNum{big: new(big.Int).Rem(n.asBig(), m.asBig()), useBig: true}
+}
```
chris-ricketts: I checked all call sites to verify zero is guarded against but wondering if it would be better to return `*BigNum` and `nil` if `m == 0` for `Div` and `Mod` - that way callers MUST check it in order to use it.

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+	return BigNum{big: new(big.Int).Mul(n.asBig(), m.asBig()), useBig: true}
+}
+
+// Div returns truncated n / m. Caller MUST verify m is non-zero first.
+// Promotes only on int64 min / -1 overflow.
+func (n BigNum) Div(m BigNum) BigNum {
+	if !n.useBig && !m.useBig {
+		if !(n.small == -9223372036854775808 && m.small == -1) {
+			return BigNum{small: n.small / m.small}
+		}
```
chris-ricketts: Nit/Lint: DeMorgan's law applied (with constant):

```suggestion
		if n.small != math.MinInt64 && m.small != -1 {
			return BigNum{small: n.small / m.small}
		}
```

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+	return BigNum{big: new(big.Int).Quo(n.asBig(), m.asBig()), useBig: true}
+}
+
+// Mod returns truncated n % m (sign follows dividend). Caller MUST verify
+// m is non-zero first.
+func (n BigNum) Mod(m BigNum) BigNum {
+	if !n.useBig && !m.useBig {
+		if !(n.small == -9223372036854775808 && m.small == -1) {
+			return BigNum{small: n.small % m.small}
+		}
```
chris-ricketts: Nit/Lint: DeMorgan's law applied (with constant):

```suggestion
		if n.small != math.MinInt64 && m.small != -1 {
			return BigNum{small: n.small % m.small}
		}
```

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+	return BigNum{big: new(big.Int).Rem(n.asBig(), m.asBig()), useBig: true}
+}
+
+// Negate returns -n. Promotes on int64 min.
+func (n BigNum) Negate() BigNum {
+	if !n.useBig {
+		if n.small != -9223372036854775808 {
+			return BigNum{small: -n.small}
+		}
```
chris-ricketts: Nit: use constant for min int64

```suggestion
		if n.small != math.MaxInt64 {
			return BigNum{small: -n.small}
		}
```

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+
+// Abs returns |n|. Promotes on int64 min.
+func (n BigNum) Abs() BigNum {
+	if !n.useBig {
+		if n.small >= 0 {
+			return n
+		}
+		if n.small != -9223372036854775808 {
+			return BigNum{small: -n.small}
+		}
```
chris-ricketts: Nit: use constant for min int64

```suggestion
		if n.small != math.MinInt64 {
			return BigNum{small: -n.small}
		}
```

pkg/arkade/asset_opcodes.go
```diff
@@ -303,24 +306,32 @@ func opcodeInspectAssetGroupSum(op *opcode, data []byte, vm *Engine) error {
...
+			return err
+		}
 		outSum := safeSumOutputs(group.Outputs)
 		if !outSum.IsUint64() {
 			return scriptError(txscript.ErrInvalidStackOperation, "amount overflow")
 		}
-		pushAmountLE64(vm, outSum.Uint64())
+		if err := vm.dstack.PushBigNum(BigNumFromUint64(outSum.Uint64())); err != nil {
+			return err
+		}
```
chris-ricketts: `safeSumInputs` and `safeSumOutputs` already return a `big.Int`.

Do we still require the `IsUint64` check now we have `BigNum`?

This could cause some confusion as technically an amount of `math.MaxUint64` is legal.

pkg/arkade/bignum_test.go
```diff
@@ -0,0 +1,396 @@
...
+		{"zero empty", nil, 0, false, "", 0},
+		{"one", []byte{0x01}, 1, false, "", 0},
+		{"neg one", []byte{0x81}, -1, false, "", 0},
+		{"127", []byte{0x7f}, 127, false, "", 0},
+		{"128", []byte{0x80, 0x00}, 128, false, "", 0},
+		{"neg 128", []byte{0x80, 0x80}, -128, false, "", 0},
+		// int64 boundary (max positive and min negative that fit in 8
+		// minimally-encoded bytes)
+		{"int64 max 8 bytes", []byte{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x7f}, 9223372036854775807, false, "", 0},
+		{"int64 min plus one 8 bytes", []byte{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff}, -9223372036854775807, false, "", 0},
```
chris-ricketts: Nit: use constants 

```suggestion
		{"int64 max 8 bytes", []byte{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x7f}, math.MaxInt64, false, "", 0},
		{"int64 min plus one 8 bytes", []byte{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff}, math.MinInt64 + 1, false, "", 0},
```

pkg/arkade/bignum_test.go
```diff
@@ -0,0 +1,396 @@
...
+		{"zero int64", BigNumFromInt64(0), nil},
+		{"one", BigNumFromInt64(1), []byte{0x01}},
+		{"neg one", BigNumFromInt64(-1), []byte{0x81}},
+		{"127", BigNumFromInt64(127), []byte{0x7f}},
+		{"128", BigNumFromInt64(128), []byte{0x80, 0x00}},
+		{"neg 128", BigNumFromInt64(-128), []byte{0x80, 0x80}},
+		{"255", BigNumFromInt64(255), []byte{0xff, 0x00}},
+		{"neg 255", BigNumFromInt64(-255), []byte{0xff, 0x80}},
+		{"int64 max", BigNumFromInt64(9223372036854775807), []byte{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x7f}},
+		{"int64 min plus one", BigNumFromInt64(-9223372036854775807), []byte{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff}},
```
chris-ricketts: Nit: use constants

```suggestion
		{"int64 max", BigNumFromInt64(math.MaxInt64), []byte{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x7f}},
		{"int64 min plus one", BigNumFromInt64(math.MinInt64 + 1), []byte{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff}},
```

pkg/arkade/bignum_test.go
```diff
@@ -0,0 +1,396 @@
...
+}
+
+func TestBigNumSubOverflowPromotes(t *testing.T) {
+	t.Parallel()
+	a := BigNumFromInt64(-9223372036854775808) // int64 min
+	b := BigNumFromInt64(1)
+	got := a.Sub(b)
+	if !got.useBig {
+		t.Fatalf("expected promotion, got %+v", got)
+	}
```
chris-ricketts: Nit: use constants

Also: add other subtraction overflow boundary case and value checks

```suggestion
	a := BigNumFromInt64(math.MinInt64)
	b := BigNumFromInt64(1)
	got := a.Sub(b)
	if !got.useBig {
		t.Fatalf("expected promotion, got %+v", got)
	}
	want := new(big.Int).Sub(big.NewInt(math.MinInt64), big.NewInt(1))
	if want.Cmp(got.big) != 0 {
		t.Fatalf("expected %d, got %d", want, got.big)
	}
	a = BigNumFromInt64(math.MaxInt64)
	b = BigNumFromInt64(-1)
	got = a.Sub(b)
	if !got.useBig {
		t.Fatalf("expected promotion, got %+v", got)
	}
	want = new(big.Int).Add(big.NewInt(math.MaxInt64), big.NewInt(1))
	if want.Cmp(got.big) != 0 {
		t.Fatalf("expected %d, got %d", want, got.big)
	}
```

pkg/arkade/bignum_test.go
```diff
@@ -0,0 +1,396 @@
...
+	q = BigNumFromInt64(7).Div(BigNumFromInt64(-2))
+	r = BigNumFromInt64(7).Mod(BigNumFromInt64(-2))
+	if q.small != -3 || r.small != 1 {
+		t.Fatalf("got q=%d r=%d, want q=-3 r=1", q.small, r.small)
+	}
+}
+
+func TestBigNumNegateOverflowPromotes(t *testing.T) {
+	t.Parallel()
+	a := BigNumFromInt64(-9223372036854775808) // int64 min; -min overflows
```
chris-ricketts: Nit: use constant

```suggestion
	a := BigNumFromInt64(math.MinInt64) // int64 min; -min overflows
```

pkg/arkade/bignum_test.go
```diff
@@ -0,0 +1,396 @@
...
+	got := a.Negate()
+	if !got.useBig {
+		t.Fatalf("expected promotion, got %+v", got)
+	}
+	want := new(big.Int).Neg(big.NewInt(-9223372036854775808))
```
chris-ricketts: Nit: use constant

```suggestion
	want := new(big.Int).Neg(big.NewInt(math.MinInt64))
```

pkg/arkade/bignum_test.go
```diff
@@ -0,0 +1,396 @@
...
+	}
+}
+
+func TestBigNumAbs(t *testing.T) {
+	t.Parallel()
+	if BigNumFromInt64(-5).Abs().small != 5 {
+		t.Fatalf("abs(-5) wrong")
+	}
+	// abs of int64 min must promote.
+	got := BigNumFromInt64(-9223372036854775808).Abs()
```
chris-ricketts: Nit: use constant 

```suggestion
	got := BigNumFromInt64(math.MinInt64).Abs()
```

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+	if len(v) <= int64ByteCap {
+		return decodeInt64(v), nil
+	}
+	return decodeBig(v), nil
+}
+
+// decodeInt64 parses up to 8 bytes of sign-magnitude LE into int64.
+// Pre: 1 ≤ len(v) ≤ 8.
+func decodeInt64(v []byte) BigNum {
```
chris-ricketts: What is the reason for not handling `len(v) == 0`?

I would have expected an encoding roundtrip to work for any `int64`.

`decodeInt64(encodeInt64(0))` panics with:

`panic: runtime error: index out of range [-1]` in `pkg/arkade/bignum.go:7`

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+	neg := v < 0
+	if neg {
+		v = -v
+	}
+	result := make([]byte, 0, 9)
+	for v > 0 {
+		result = append(result, byte(v&0xff))
+		v >>= 8
+	}
+	if result[len(result)-1]&0x80 != 0 {
```
chris-ricketts: `encodeInt64(math.MinInt64)` panics on this line with: ` index out of range [-1]`

pkg/arkade/bignum_test.go
```diff
@@ -0,0 +1,396 @@
...
+			if err != nil {
+				t.Fatalf("Bytes() error: %v", err)
+			}
+			if !bytes.Equal(got, tc.want) {
+				t.Fatalf("Bytes() = %x, want %x", got, tc.want)
+			}
+		})
+	}
+}
+
```
chris-ricketts, line 127: Add an encoding roundtrip test for `int64`:

```suggestion
func TestInt64EncodingRoundtrip(t *testing.T) {
	t.Parallel()

	got := decodeInt64(encodeInt64(math.MaxInt64))
	if got.useBig {
		t.Fatalf("expected no promotion to big, got %+v", got)
	}
	if got.small != math.MaxInt64 {
		t.Fatalf("expected %d, got %d", math.MaxInt64, got.small)
	}

	got = decodeInt64(encodeInt64(math.MinInt64))
	if got.useBig {
		t.Fatalf("expected no promotion to big, got %+v", got)
	}
	if got.small != math.MinInt64 {
		t.Fatalf("expected %d, got %d", math.MaxInt64, got.small)
	}

	got = decodeInt64(encodeInt64(0))
	if got.useBig {
		t.Fatalf("expected no promotion to big, got %+v", got)
	}
	if got.small != 0 {
		t.Fatalf("expected %d, got %d", 0, got.small)
	}
}
```

pkg/arkade/bignum_test.go
```diff
@@ -0,0 +1,396 @@
...
+	got, err = neg.Bytes()
+	if err != nil {
+		t.Fatalf("Bytes: %v", err)
+	}
+	want = []byte{0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x80, 0x80}
+	if !bytes.Equal(got, want) {
+		t.Fatalf("Bytes() = %x, want %x", got, want)
+	}
+}
+
```
chris-ricketts, line 242: Add `int.Big` encoding round trip test with boundary cases:

```suggestion
func TestBigEncodingRoundtrip(t *testing.T) {
	t.Parallel()
	tests := []struct {
		name string
		src  *big.Int
	}{
		{
			name: "2^(512*8)",
			src:  new(big.Int).Lsh(big.NewInt(1), 512*8),
		},
		{
			name: "-2^(512*8)",
			src: new(big.Int).Neg(
				new(big.Int).Lsh(big.NewInt(1), 512*8),
			),
		},
		{
			name: "max int64 plus one",
			src: new(big.Int).Add(
				big.NewInt(math.MaxInt64),
				big.NewInt(1),
			),
		},
		{
			name: "min int64 minus one",
			src: new(big.Int).Sub(
				big.NewInt(math.MinInt64),
				big.NewInt(1),
			),
		},
	}
	for _, tc := range tests {
		t.Run(tc.name, func(t *testing.T) {
			enc := encodeBig(tc.src)
			if len(enc) < 9 {
				t.Fatalf("encodeBig() length = %d, want at least 9", len(enc))
			}
			got := decodeBig(enc)
			if !got.useBig {
				t.Fatalf("expected big-path result, got %+v", got)
			}
			if got.big == nil {
				t.Fatalf("expected non-nil big.Int")
			}
			if got.big.Cmp(tc.src) != 0 {
				t.Fatalf("roundtrip mismatch: got %s, want %s", got.big, tc.src)
			}
		})
	}
}
```

pkg/arkade/bignum_test.go
```diff
@@ -0,0 +1,396 @@
...
+func TestBigNumAddOverflowPromotes(t *testing.T) {
+	t.Parallel()
+	a := BigNumFromInt64(9223372036854775807) // int64 max
+	b := BigNumFromInt64(1)
+	got := a.Add(b)
+	if !got.useBig {
+		t.Fatalf("expected promotion to big, got %+v", got)
+	}
+	want := new(big.Int).Add(big.NewInt(9223372036854775807), big.NewInt(1))
```
chris-ricketts: Nit: use constants

```suggestion
	a := BigNumFromInt64(math.MaxInt64) // int64 max
	b := BigNumFromInt64(1)
	got := a.Add(b)
	if !got.useBig {
		t.Fatalf("expected promotion to big, got %+v", got)
	}
	want := new(big.Int).Add(big.NewInt(math.MaxInt64), big.NewInt(1))
```

pkg/arkade/opcode.go
```diff
@@ -948,103 +946,58 @@ func verifyLockTime(txLockTime, threshold, lockTime int64) error {
...
-
-	// In the rare event that the argument needs to be < 0 due to some
-	// arithmetic being done first, you can always use
-	// 0 OP_MAX OP_CHECKLOCKTIMEVERIFY.
-	if lockTime < 0 {
-		str := fmt.Sprintf("negative lock time: %d", lockTime)
-		return scriptError(txscript.ErrNegativeLockTime, str)
+	// Locktime must fit in uint32. Values on the big path are ≥ 2^63,
+	// which cannot be a valid locktime.
+	if n.useBig || n.small > int64(math.MaxUint32) {
```
chris-ricketts: Nit: no need to cast untyped int constants.

```suggestion
	if n.useBig || n.small > math.MaxUint32 {
```

pkg/arkade/opcode.go
```diff
@@ -948,103 +946,58 @@ func verifyLockTime(txLockTime, threshold, lockTime int64) error {
...
+			fmt.Sprintf("negative sequence: %s", n.asBig().Text(10)))
 	}
-
-	// In the rare event that the argument needs to be < 0 due to some
-	// arithmetic being done first, you can always use
-	// 0 OP_MAX OP_CHECKSEQUENCEVERIFY.
-	if stackSequence < 0 {
-		str := fmt.Sprintf("negative sequence: %d", stackSequence)
-		return scriptError(txscript.ErrNegativeLockTime, str)
+	if n.useBig || n.small > int64(math.MaxUint32) {
```
chris-ricketts: Nit: unnecessary untyped integer const cast

```suggestion
	if n.useBig || n.small > math.MaxUint32 {
```

pkg/arkade/opcode.go
```diff
@@ -1721,26 +1587,85 @@ func opcodeMax(op *opcode, data []byte, vm *Engine) error {
...
+	sign := encoded[len(encoded)-1] & 0x80
+	magnitude := append([]byte(nil), encoded...)
+	magnitude[len(magnitude)-1] &= 0x7f
+	if magnitude[len(magnitude)-1] == 0 {
+		magnitude = magnitude[:len(magnitude)-1]
+	}
+
+	copy(out, magnitude)
+	out[len(out)-1] |= sign
+	vm.dstack.PushByteArray(out)
```
chris-ricketts: Maybe this fiddly logic should be moved to a method like `FixedBytes` on `BigNum` rather than live in the opcode handler. wdyt?

test/counter_contract_test.go
```diff
@@ -257,7 +256,14 @@ func addCounterPacket(t *testing.T, ptx *psbt.Packet, value uint64) {
...
-	return uint64LE(value)
+	// The extension packet format cannot represent an empty data field, while
+	// zero minimally encodes as empty. Store the test counter offset by one so
+	// every payload remains a canonical BigNum and can feed OP_ADD directly.
+	payload, err := arkade.BigNumFromUint64(value + 1).Bytes()
+	if err != nil {
+		panic(err)
+	}
+	return payload
 }
```
louisinger: let's avoid panic in tests. instead you should pass testing.T and use require.NoError(t, err)

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+	if len(out) > maxBigNumLen {
+		return nil, scriptError(txscript.ErrNumberTooBig,
+			fmt.Sprintf("BigNum result encoded as %d bytes exceeds max allowed of %d",
+				len(out), maxBigNumLen))
+	}
+	return out, nil
+}
+
+// encodeInt64 reproduces the legacy scriptNum.Bytes() algorithm for int64.
+func encodeInt64(v int64) []byte {
```
louisinger: move unexported functions at the bottom of the file. and all exported ones before. especially the BigNum methods should be declared next to the struct.

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+	// toward negative infinity for negative values, matching our spec.
+	res := new(big.Int).Rsh(n.asBig(), shift)
+	return BigNum{big: res, useBig: true}
+}
+
+// MinimallyEncode returns the minimal sign-magnitude LE encoding of the
+// byte slice v (interpreting v as sign-magnitude LE). It strips trailing
+// zero-bytes while preserving the sign bit, and normalises negative zero
+// ([0x80], [0x00, 0x00, 0x80], etc.) to the empty slice.
+func MinimallyEncode(v []byte) []byte {
```
louisinger: It really has to be exported ? it seems only used internally by the engine. do u see any case where the user building the script will use it ? if no, let's unexport it to reduce API surface.

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+// MakeBigNum decodes a sign-magnitude little-endian byte slice into a BigNum.
+// If requireMinimal is true, inputs that are not minimally encoded are
+// rejected (including the negative-zero encoding [0x80]). If len(v) > maxLen
+// an ErrNumberTooBig is returned.
+//
+// Values with len(v) ≤ 8 land on the int64 fast path; ≥ 9 bytes land on the
+// big.Int path.
+func MakeBigNum(v []byte, requireMinimal bool, maxLen int) (BigNum, error) {
```
louisinger:
```suggestion
func BigNumFromBytes(v []byte) (BigNum, error) {
```
ideally I want to see this API (simple byte slice parameter + name close to other factory functions).

I know the current signature is inspired by the bitcoin "scriptNum" but from developer point of view "requireMinimal" and "maxLen" is something u don't really understand if u're not a script wizard. can we simplify it ? 

the use case I have in mind is : the dev decodes the script, see some bytes slice, call a simpler `BigNumFromBytes` and get a number

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+		case n.small > m.small:
+			return 1
+		}
+		return 0
+	}
+	return n.asBig().Cmp(m.asBig())
+}
+
+// asBig materialises a *big.Int view of n regardless of current path.
+func (n BigNum) asBig() *big.Int {
```
louisinger:
```suggestion
func (n BigNum) BigInt() *big.Int {
```
rename + export (can be useful if the dev decodes some script and want to do pure golang arithmetic)

pkg/arkade/bignum.go
louisinger: it seems to me we can admit `requireMinimal` is always true and `maxLen` seems to be a constant

louisinger (changes requested):

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+func decodeBig(v []byte) BigNum {
+	msb := v[len(v)-1]
+	negative := msb&0x80 != 0
+	mag := make([]byte, len(v))
+	copy(mag, v)
+	mag[len(mag)-1] = msb & 0x7f
+	// Reverse to big-endian for big.Int.SetBytes.
+	for i, j := 0, len(mag)-1; i < j; i, j = i+1, j-1 {
+		mag[i], mag[j] = mag[j], mag[i]
+	}
```
louisinger: use `slices.Reverse(mag)` instead of custom for loop

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+// ([0x80], [0x00, 0x00, 0x80], etc.) to the empty slice.
+func MinimallyEncode(v []byte) []byte {
+	if len(v) == 0 {
+		return []byte{}
+	}
+	out := append([]byte(nil), v...)
+	sign := out[len(out)-1] & 0x80
+	// Clear sign bit from MSB so we can detect an all-zero magnitude.
+	out[len(out)-1] &= 0x7f
+	// Strip trailing zero-bytes.
```
louisinger: to remove trailing zero : `bytes.TrimRight(out, "\x00")`

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+// encodeBig serialises a *big.Int as minimal sign-magnitude LE.
+func encodeBig(v *big.Int) []byte {
+	if v.Sign() == 0 {
+		return nil
+	}
+	mag := new(big.Int).Abs(v).Bytes() // big-endian magnitude
+	le := make([]byte, len(mag))
+	for i, b := range mag {
+		le[len(mag)-1-i] = b
+	}
```
louisinger: use `bytes.Reverse`

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+		switch {
+		case n.small < 0:
+			return -1
+		case n.small > 0:
+			return 1
+		}
+		return 0
+	}
+	return n.big.Sign()
+}
```
louisinger, line 103:
```suggestion
// Sign returns -1, 0, or +1.
  func (n BigNum) Sign() int {
      if n.useBig {
          return n.big.Sign()
      }                                                                                                                                                                                       
      return cmp.Compare(n.small, int64(0))
  }
```
we can simplify using builtin "cmp"

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,356 @@
...
+		switch {
+		case n.small < m.small:
+			return -1
+		case n.small > m.small:
+			return 1
+		}
+		return 0
+	}
+	return n.asBig().Cmp(m.asBig())
+}
```
louisinger, line 120:
```suggestion
// Cmp reports -1/0/+1 comparing n and m.
func (n BigNum) Cmp(m BigNum) int {                                                                                                                                                         
    if n.useBig || m.useBig {
        return n.asBig().Cmp(m.asBig())                                                                                                                                                     
    }                                                              
    return cmp.Compare(n.small, m.small)
}
```
let's also use built-in "cmp" golang pkg

Kukks:
@arkanaai Review

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,417 @@
...
+// BigNum operand or result. Equal to txscript.MaxScriptElementSize.
+const maxBigNumLen = txscript.MaxScriptElementSize
+
+// int64ByteCap is the largest byte length whose minimal sign-magnitude LE
+// encoding is guaranteed to fit in int64. 9+ bytes require the big.Int path.
+const int64ByteCap = 8
+
+var (
+	errBigNumDivisionByZero = errors.New("division by zero")
+	errBigNumModuloByZero   = errors.New("modulo by zero")
```
louisinger:
```suggestion
	ErrBigNumDivisionByZero = errors.New("division by zero")
	ErrBigNumModuloByZero   = errors.New("modulo by zero")
```
exports to let `Div` and `Mod` user reuse it just in case

louisinger (approved):
small nit about the defined errors but LGTM

edit: BigInt is at the bottom, probably a leftover

pkg/arkade/bignum.go
```diff
@@ -0,0 +1,417 @@
...
+		le = append(le, extra)
+	} else if neg {
+		le[len(le)-1] |= 0x80
+	}
+	return le
+}
+
+// BigInt returns n as a fresh *big.Int. The returned value is independent of
+// n's internal state; callers may mutate it freely.
+func (n BigNum) BigInt() *big.Int {
```
louisinger: move close to other methods

louisinger (changes requested):

chris-ricketts (commented):
LGTM - overall great simplification and ergonomic improvement

PR #74, opened 2026-04-24
test: non interactive delegate

louisinger:
add a test for a covenant delegating vtxo refresh to any solver. refreshing is safe from the user but doens't require its signature.

@Kukks @msinkec please review

### DIFF
--- a/test/delegate_test.go
+++ b/test/delegate_test.go
@@ -1,0 +1,427 @@
+package test
+
+import (
+	"bytes"
+	"context"
+	"encoding/hex"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/ArkLabsHQ/introspector/pkg/arkade"
+	introspectorclient "github.com/ArkLabsHQ/introspector/pkg/client"
+	arklib "github.com/arkade-os/arkd/pkg/ark-lib"
+	"github.com/arkade-os/arkd/pkg/ark-lib/intent"
+	"github.com/arkade-os/arkd/pkg/ark-lib/offchain"
+	"github.com/arkade-os/arkd/pkg/ark-lib/script"
+	"github.com/arkade-os/arkd/pkg/ark-lib/tree"
+	"github.com/arkade-os/arkd/pkg/ark-lib/txutils"
+	arksdk "github.com/arkade-os/go-sdk"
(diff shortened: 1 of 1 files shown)


test/delegate_test.go
```diff
@@ -0,0 +1,427 @@
...
+		{Value: delegateAmount - 1, PkScript: delegatePkScript},
+	})
+
+	// Invalid: off-chain Ark tx (v3) rejected by the version gate
+	infos, err := grpcAlice.GetInfo(ctx)
+	require.NoError(t, err)
+	checkpointScriptBytes, err := hex.DecodeString(infos.CheckpointTapscript)
+	require.NoError(t, err)
+
+	offchainPtx, offchainCheckpoints, err := offchain.BuildTxs(
```
msinkec, line 218: Maybe set PrevArkTxField for robustness, so the script passes if we make a regression on OP_INSPECTVERSION?

msinkec (approved):
LGTM 🤝

test/delegate_test.go
louisinger: wdym ? not sure I get it

msinkec: I meant after 
```
 addIntrospectorPacket(t, offchainPtx, []arkade.IntrospectorEntry{
  	{Vin: 0, Script: delegateArkadeScript},
  })
```
add
```
  require.NoError(t, txutils.SetArkPsbtField(
  	offchainPtx, 0, arkade.PrevArkTxField, *fundingTx,
  ))
```

OP_INSPECTINPUTSCRIPTPUBKEY does not use the checkpoint prevout directly. Itt needs PrevArkTxField so the introspector can recover the original VTXO scriptPubKey behind the checkpoint. Today the version check fails first, so the missing field is not observed. Adding it makes the v3 case fully valid except for OP_INSPECTVERSION, which is exactly what the test wants to prove.

It's just to make the test more robust for regression. Just a nitpick.

louisinger: this covenant aims only to run in batch swap so it will never run for ark tx (with checkpoints). that's why not needed here. 
it's to delegate the renewal of the coin so u want to explicitly disallow ppl to use it "offchain" wil classic ark tx because a malicious attack could use it to create a really long chain of transactions, making the exit super expensive.

PR #75, opened 2026-04-27
test: asset account covenant

louisinger:
@msinkec @Kukks please review


msinkec:
Liveness depends on the solver and Alices funds can get stuck in the covenant. Alice needs an additional timeout path to reclaim funds in such a case.

PR #79, opened 2026-05-14
Implement OP_ECADD, OP_ECMUL, OP_ECPAIRING

msinkec:
## Summary

Adds three Arkade Script opcodes for elliptic curve arithmetic, selectable at runtime by `curve_id`:

| Word           | Opcode | Hex    | Stack signature                                                    | Output  |
| -------------- | ------ | ------ | ------------------------------------------------------------------ | ------- |
| `OP_ECADD`     | 246    | `0xf6` | `x1 y1 x2 y2 curve_id`                                             | `x3 y3` |
| `OP_ECMUL`     | 247    | `0xf7` | `x y k curve_id`                                                   | `x2 y2` |
| `OP_ECPAIRING` | 248    | `0xf8` | `[g1_x g1_y g2_x_c1 g2_x_c0 g2_y_c1 g2_y_c0]… pair_count curve_id` | `bool`  |

Curves at launch:

| Curve ID | Curve                    | Operations        |
| -------- | ------------------------ | ----------------- |
| 0        | `secp256k1`              | add, mul          |
| 1        | `secp256r1` (NIST P-256) | add, mul          |
| 2        | `alt_bn128` / BN254      | add, mul, pairing |

`0xf9` stays reserved. Pairing only works for `alt_bn128` at launch; the curve_id argument is kept so a future BLS12 curve can reuse the same opcode shape.

## Stack interface

All numeric inputs and outputs — curve IDs, pair counts, affine coordinates, Fp2 components, scalars, result coordinates — are Arkade BigNums (minimally encoded sign-magnitude little-endian). A script can build a coordinate with `OP_ADD`/`OP_SUB` and feed it directly into `OP_ECADD` without an extra fixed-width encoding step.

- `(0, 0)` is the wire representation of the G1 point at infinity. G2 infinity is all four `Fp2` components zero.
- `OP_ECPAIRING` follows EIP-197 coordinate order for `alt_bn128` G2: each pair is six BigNums `(g1_x, g1_y, g2_x_c1, g2_x_c0, g2_y_c1, g2_y_c0)`.
- `k = 0` for `OP_ECMUL` returns infinity. `pair_count = 0` for `OP_ECPAIRING` returns true (matches EIP-197).
- Scalars must be `0 ≤ k < group_order`. Coordinates must be `0 ≤ x < field_modulus`. This intentionally differs from EIP-196, where the scalar is a 32-byte calldata blob that can exceed the group order — Arkade Script has native arbitrary-precision arithmetic, so we reject non-canonical scalars at the opcode boundary.

## Validation and failure model

Every structural or value error fails script execution; only a valid pairing-product that happens to evaluate to a non-identity element pushes canonical false. Failure cases:

- stack underflow
- unsupported `curve_id`
- unsupported pairing curve (only `alt_bn128` supported at launch)
- non-minimal BigNum inputs
- negative coordinates, field elements, scalars, or pair counts
- coordinates `≥ field_modulus`
- scalars `≥ group_order`
- off-curve G1 / G2
- G2 points outside the `alt_bn128` r-subgroup (explicit subgroup check; G1 has cofactor 1 so on-curve suffices)
- `pair_count > 16`

Errors map to `txscript.ErrInvalidStackOperation` except where existing stack/BigNum helpers already produce a more specific code (`ErrMinimalData`, `ErrNumberTooBig`, stack-underflow `ErrInvalidStackOperation`).

`pair_count` is hard-capped at 16. Arkade Script has no gas model, so pairing CPU work must be bounded deterministically.

## Implementation

All curve logic and adapter code lives in `pkg/arkade/ec_ops.go`:

- Curve metadata (id, name, field modulus, group order, `supportsPairing`) in one table.
- BigNum-to-`*big.Int` validation helpers at the opcode boundary (`popCurveID`, `popInFieldElement`, `popInGroupScalar`).
- Three opcode handlers: `opcodeECAdd`, `opcodeECMul`, `opcodeECPairing`.
- Per-curve adapters that hide library types from the handlers and convert outputs back to canonical BigNums.

Backends:

- `alt_bn128` (G1, G2, pairing) and `secp256k1` (G1) — `github.com/consensys/gnark-crypto` v0.19.2 (pinned).
- `secp256r1` / NIST P-256 — Go standard library `crypto/elliptic.P256()` (the same `crypto/internal/nistec` implementation used by `crypto/tls`, `crypto/ecdsa`, and `crypto/ecdh`).


ghost (changes requested):
## 🔍 Arkana Code Review — OP_ECADD / OP_ECMUL / OP_ECPAIRING

**Verdict: Request Changes** — protocol-critical consensus opcodes require human sign-off before merge. The implementation is high quality overall, but there are issues to address and downstream coordination required.

---

### ⚠️ PROTOCOL-CRITICAL — REQUIRES HUMAN REVIEW

This PR adds **three new consensus opcodes** to the Arkade Script VM. Any implementation divergence between nodes causes chain forks. This must not merge without explicit human sign-off from protocol leads.

---

### Cross-Repo Breakage — Action Required

**1. `rust-sdk` must be updated (blocking)**
`ark-script/src/opcodes.rs` defines `ARKADE_OPCODES` ending at `0xf5`. The comment reads:
> "Arkade opcodes occupy byte values `0xb3` and `0xc4..=0xf5`"

Any Rust consumer using `ARKADE_OPCODES` to validate scripts will reject `0xf6`–`0xf8` as unknown. This needs a coordinated update — ideally an open PR on `rust-sdk` before this merges.

**2. Downstream Go consumers**: `bancod` and `layerzero-usdt0-arkade-demo` both import `github.com/ArkLabsHQ/introspector/pkg/arkade`. They will pick up these opcodes on next `go get`. No breakage expected (new opcodes only, no API changes), but worth a heads-up.

---

### Findings

#### 1. [Medium] No activation gate for new opcodes
`pkg/arkade/opcode.go:609-611` — The three opcodes go from `opcodeInvalid` → live handlers with no version/height/flag gate. If any node on the network is running the old code when a script uses `OP_ECADD`, it will reject the transaction as invalid while updated nodes accept it — **chain split**.

Is there an activation mechanism outside this repo (e.g., in `arkd`)? If not, this needs one, or the deployment plan needs to guarantee all nodes upgrade atomically.

#### 2. [Low] `crypto/elliptic.P256()` Add/ScalarMult deprecation risk
`pkg/arkade/ec_ops.go:627-657` — The deprecated `crypto/elliptic` API is used for secp256r1. The `//nolint:staticcheck` suppression is acknowledged in the comments, which is good. However, Go could change or remove the deprecated methods in a future release. Consider documenting the minimum Go version constraint, or migrating to `crypto/ecdh` with manual point serialization if feasible.

#### 3. [Low] Missing edge-case test vectors
The test suite is solid but a few boundary vectors are missing:

- **`k = order - 1`** for each curve (last valid scalar — confirms the boundary check is `<` not `<=`)
- **`pair_count = 16`** (exactly at `maxECPairingCount` — confirms the boundary is `>` not `>=`)
- **`OP_ECMUL` with `secp256r1` infinity input** (`P=(0,0), k>0`) — tested for secp256k1 but not P-256
- **`OP_ECADD` with `secp256r1` infinity + infinity** — tested for secp256k1 but not P-256

These are consensus-critical boundaries — a fence-post error here means valid scripts fail or invalid scripts succeed.

#### 4. [Info] Pairing CPU bound
`maxECPairingCount = 16` is reasonable. BN254 pairing-product check with 16 pairs on gnark-crypto should complete in ~50-100ms. Worth benchmarking and documenting the expected worst-case latency to justify the cap in the spec.

---

### What Looks Good ✅

- **On-curve validation** on all inputs before computation — prevents invalid-point attacks
- **G2 subgroup check** (`IsInSubGroup()`) for BN254 pairing — critical for soundness, correctly implemented
- **Field/scalar bounds checking** — `popInFieldElement` enforces `0 ≤ n < modulus`, `popInGroupScalar` enforces `0 ≤ k < order`
- **Minimal BigNum encoding** enforced via existing `PopBigNum` — no non-canonical encoding ambiguity
- **Infinity point handling** — `(0,0)` convention consistently applied across all three curves with explicit special-casing for P-256's `crypto/elliptic` which rejects `(0,0)` from `IsOnCurve`
- **Fuzz test builders** (`ecCurveCaseBuilder`, `ecPairingFuzzCaseBuilder`) — well-designed to penetrate past validation gates
- **gnark-crypto v0.19.2 pinned** — correct for consensus-critical code
- **Error codes** — all failures map to documented consensus error codes
- **Stack discipline** — pop/push order verified against spec; alt-stack and cond-stack invariants checked in property tests
- **Pairing follows EIP-197 semantics** — empty set returns true, coordinate order matches, but with stricter scalar validation (good deviation from EIP-196)
- **Comprehensive test coverage** — off-curve, out-of-field, negative values, non-minimal encoding, subgroup violations all tested

---

### Summary

The EC arithmetic implementation is well-engineered and the security-critical checks (on-curve, subgroup, field bounds) are all present. The main concerns are:

1. **Activation gating** — how do we prevent chain splits during rollout?
2. **rust-sdk coordination** — needs `0xf6-0xf8` added to `ARKADE_OPCODES`
3. **Missing boundary test vectors** — `k = order-1` and `pair_count = 16`

Once the activation strategy is clarified and the missing tests are added, this is ready for human protocol-lead sign-off.

---
*🤖 Reviewed by Arkana · aggressive code review · model: opus*

README.md
```diff
@@ -317,8 +317,20 @@ and can be up to the maximum script element size. `OP_NUM2BIN` and
...
+| OP_ECMULSCALARVERIFY | 227 | 0xe3 | k P Q | Nothing/fail | Verifies that Q = k*P on secp256k1 where k is a 32-byte scalar, P is a compressed public key, and Q is a compressed public key. Fails if verification fails. |
 | OP_TWEAKVERIFY | 228 | 0xe4 | P k Q | Nothing/fail | Verifies that Q = P + k*G where P is a 32-byte X-only internal key, k is a 32-byte big-endian scalar, Q is a 33-byte compressed point, and G is the generator point. Fails if verification fails. |
+| OP_ECADD | 246 | 0xf6 | x1 y1 x2 y2 curve_id | x3 y3 | Adds two affine points on the selected curve. Coordinates are Arkade BigNums; `(0, 0)` represents the point at infinity. Fails on unsupported `curve_id`, non-minimal BigNums, negative or out-of-field coordinates, and off-curve points. |
+| OP_ECMUL | 247 | 0xf7 | x y k curve_id | x2 y2 | Multiplies an affine point by a scalar on the selected curve. `k = 0` returns the point at infinity. Fails on scalars `>=` the group order, off-curve points, and the other validation cases listed for OP_ECADD. |
```
louisinger: why (x,y) affine coords instead of compressed public key (33 bytes) like OP_ECMULSCALARVERIFY ? make sense to have coords because it's low-level primitive but we should maybe uniformize. We can consider dropping OP_ECMULSCALARVERIFY and OP_TWEAKVERIFY (it comes from liquid)

pkg/arkade/opcode.go
```diff
@@ -305,9 +305,9 @@ const (
 	OP_TXID                          = 0xf3 // 243
 	OP_INSPECTPACKET                 = 0xf4 // 244
 	OP_INSPECTINPUTPACKET            = 0xf5 // 245
-	OP_UNKNOWN246                    = 0xf6 // 246
-	OP_UNKNOWN247                    = 0xf7 // 247
-	OP_UNKNOWN248                    = 0xf8 // 248
+	OP_ECADD                         = 0xf6 // 246
+	OP_ECMUL                         = 0xf7 // 247
+	OP_ECPAIRING                     = 0xf8 // 248
```
louisinger: nit: let's move EC opcodes close to `OP_ECMULSCALARVERIFY` and `OP_TWEAKVERIFY` ?

louisinger (commented):
concept ACK but we need a strategy for the liquid opcodes because they are confusing. I have the feeling we can "drop" them and replace by these new primitives ?

README.md
msinkec: You want to have them in an affine form if you want to use these opcodes as building blocks for some zk protocol script for example. I assume OP_ECMULSCALARVERIFY and OP_TWEAKVERIFY were used to derive bitcoin specific keys / addresses in script, hence operating on the compressed form. It's a different use case. Maybe they're still worth keeping since they could still come in handy at some point?

Techincally you can emulate both of them with the new opcodes now but you would have to do compression / decompression within script which I assume would be very awkward, since I believe it uses big endian representation under the hood. In that case it would make sense to also introduce new opcodes for it, e.g.:
```
OP_ECCOMPRESS    [x y curve_id] -> [compressed_point]
OP_ECDECOMPRESS  [compessed_point curve_id] -> [x y]
```
or you could introduce more general operations such as OP_NUM2BIN_BE + OP_BIN2NUM_BE and OP_REVERSEBYTES, which would also be useful for other stuff.

---

I'd say we keep the two old opcodes for now and we can always add the above opcodes in the future.

louisinger: > I'd say we keep the two old opcodes for now and we can always add the above opcodes in the future.

makes sense

louisinger (approved):

PR #84, opened 2026-05-20
Implement OP_REVERSEBYTES

msinkec:
## Summary
- Add `OP_REVERSEBYTES` (`0xd9`) — pops the top stack item and pushes its bytes in reverse order. 
- Repurposes the previously-unused `OP_UNKNOWN217` slot.

## Notes
- Operates on opaque bytes (no `scriptNum`/`BigNum` decoding). Length is preserved, so the existing `MaxScriptElementSize` stack limit covers the output without an explicit check.
- Empty input → empty output; 1-byte input is unchanged.
- Stack underflow surfaces as `ErrInvalidStackOperation`.


ghost (changes requested):
## Review: Implement OP_REVERSEBYTES — **Request Changes**

### 🔴 Bug: In-place mutation of shared stack buffer

**`pkg/arkade/opcode.go:1660-1666`** — `opcodeReverseBytes` mutates the popped buffer in-place with `slices.Reverse(buf)`. This violates the stack's immutability contract.

From `stack.go:32-35`:
```go
// stack represents a stack of immutable objects to be used with bitcoin
// scripts.  Objects may be shared, therefore in usage if a value is to be
// changed it *must* be deep-copied first to avoid changing other values on
// the stack.
```

`PopByteArray()` → `nipN(0)` returns a **direct reference** to the internal `[]byte` (stack.go:188). If the same element was duplicated via `OP_DUP`/`OP_OVER`/`OP_2DUP`/`OP_PICK` before `OP_REVERSEBYTES`, the duplicate still on the stack shares the same backing array. `slices.Reverse(buf)` silently corrupts it.

**Exploit scenario:**
```
OP_DUP OP_REVERSEBYTES
```
Expected stack: `[original, reversed]`. Actual stack: `[reversed, reversed]` — both entries point to the same mutated `[]byte`.

In a script that checks equality of the original against a hash (e.g., in a covenant), this could cause a valid spend to fail or an invalid spend to succeed depending on the script logic.

**Fix:**
```go
func opcodeReverseBytes(op *opcode, data []byte, vm *Engine) error {
    buf, err := vm.dstack.PopByteArray()
    if err != nil {
        return err
    }

    result := make([]byte, len(buf))
    for i, b := range buf {
        result[len(buf)-1-i] = b
    }
    vm.dstack.PushByteArray(result)
    return nil
}
```

This also eliminates the `slices` import if no other callsite needs it.

Every other mutating opcode in this file follows the clone-before-modify pattern:
- `opcodeInvert` (line ~2232): `result := make([]byte, len(x))`
- `opcodeAnd/Or/Xor` (line ~2249+): `result := make([]byte, len(x1))`
- `opcodeBin2Num` (line ~1638): `minimallyEncode()` clones with `append([]byte(nil), v...)`

### 🟡 Missing test: shared-buffer scenario

The test vectors only push a single item and reverse it. Add a vector that exercises the sharing path:

```go
{
    name:          "dup_then_reverse_no_aliasing",
    inputStack:    [][]byte{{0x01, 0x02, 0x03}},
    // Prepend OP_DUP before OP_REVERSEBYTES in the script
    // Expected: bottom = {0x01, 0x02, 0x03}, top = {0x03, 0x02, 0x01}
},
```

This would have caught the bug above.

### ✅ Everything else looks correct

- **Opcode slot `0xd9`**: Clean repurpose of `OP_UNKNOWN217`. No downstream consumers reference this slot (checked bancod, all SDKs, compiler, tapscripts — zero hits).
- **opcodeArray registration**: Length 1 ✓, handler wired correctly ✓.
- **Error handling**: Stack underflow correctly surfaces via `PopByteArray` → `ErrInvalidStackOperation` ✓.
- **Size safety**: PR description correctly notes that length is preserved, so `MaxScriptElementSize` covers it without an explicit check ✓.
- **Test coverage**: Empty, 1-byte, 3-byte, 32-byte, max-size (520), max-size-1 (519), and underflow are all covered ✓. Property checker validates alt-stack/condStack invariance ✓.
- **No cross-repo breakage**: No external repo references `OP_UNKNOWN217` by name or value.

### Summary

One real bug (shared-buffer mutation) that could cause incorrect script execution. The fix is a one-line change to allocate a new buffer instead of reversing in-place. Please also add a test vector that exercises `OP_DUP OP_REVERSEBYTES` to guard against regression.

---
🤖 Reviewed by Arkana

ghost (approved):
## ✅ Arkana Re-review — LGTM

Previous findings from my initial review have all been addressed in `7451cd2f`:

1. **Shared-buffer mutation bug** — Fixed. `opcodeReverseBytes` now allocates a fresh `[]byte` via `make([]byte, len(buf))` and copies in reverse order. No in-place mutation of the popped slice. Matches the clone-before-modify pattern used by `opcodeInvert`, `opcodeAnd`, `opcodeOr`, `opcodeXor`, and `opcodeBin2Num`.

2. **Missing aliasing test** — Added. `shared_backing_buffer_not_mutated` vector pushes two references to the same `[]byte` and verifies the lower entry is not corrupted after reversal.

3. **Documentation** — Comment on `opcodeReverseBytes` explicitly references the stack.go immutability contract.

No new issues in the merge commit `3e33b2a6`.

No protocol-critical paths touched (this is a new script opcode in the introspector VM, not VTXO/signing/forfeit/round/exit logic).

louisinger (approved):

chris-ricketts (approved):

PR #86, opened 2026-05-22
feat: support both current and deprecated signing keys

### LINKED ISSUE #83: Add support for current and deprecated introspector signing keys
## Problem

Introspector currently loads a single secret signing key from environment configuration. This blocks safe key rotation because scripts created against an older introspector signing key cannot be handled once the service starts using a new key.

Operators need to configure one current signing key plus multiple deprecated signing keys. Clients also need to discover those deprecated public keys so tests and integrations can build historical-key scenarios.

Environment variables should remain the only application configuration interface for signing keys. Deployment platforms such as AWS ECS can inject values sourced from SSM Parameter Store or Secrets Manager as environment variables, so adding an introspector-level key-source selector would duplicate infrastructure concerns and create unnecessary configuration branching.

## Solution

Load signing keys from environment variables:

- One current signing key from the existing secret key env var
- Zero or more deprecated signing keys from a new strict comma-separated env var

The application service should store a current signer and deprecated signers.

For normal script execution paths, the service should resolve the signer per Arkade script input by trying the current signer first, then deprecated signers.

For finalization, the service should use a dedicated flow: inspect the signed intent proof, determine which configured signer produced or validates each proof signature, carry that signer association forward, and use the same signer for matching forfeit or commitment signing.

`GetInfo` should continue exposing the current signer public key and add deprecated signer public keys.

### User Stories

1. As an introspector operator, I want to configure a current signing key, so that the service can continue signing new Arkade flows.

2. As an introspector operator, I want to configure deprecated signing keys, so that historical Arkade scripts remain signable after key rotation.

3. As an introspector operator, I want deprecated keys validated at startup, so that bad key material fails fast.

4. As an introspector operator, I want malformed deprecated key CSV values rejected, so that deployment mistakes do not silently change signing behavior.

5. As an introspector operator, I want duplicate raw private keys rejected, so that key rotation configuration remains unambiguous.

6. As an introspector client, I want `GetInfo` to return the current signer public key, so that I can build new scripts against the active introspector identity.

7. As an introspector client, I want `GetInfo` to return deprecated signer public keys, so that I can discover historical keys accepted by the service.

8. As an introspector client, I want deprecated public keys encoded the same way as the current public key, so that client parsing remains consistent.

9. As an integrator, I want the client wrapper to expose deprecated signer public keys, so that I do not need to bypass the official client API.

10. As an Arkade flow submitter, I want offchain transaction signing to work for scripts using a deprecated introspector key, so that historical VTXOs remain spendable.

11. As an Arkade flow submitter, I want intent proof signing to work for scripts using a deprecated introspector key, so that historical intent flows remain usable.

12. As an Arkade flow submitter, I want finalization signing to work for scripts using a deprecated introspector key, so that historical batch flows can complete.

13. As an Arkade flow submitter, I want onchain transaction signing to work for scripts using a deprecated introspector key, so that historical onchain paths remain usable.

14. As an Arkade flow submitter, I want each input resolved independently, so that a transaction can include inputs tied to different configured introspector keys.

15. As an Arkade flow submitter, I want unrelated signer inputs to preserve existing skip/fail behavior, so that key rotation does not change endpoint semantics.

16. As an Arkade flow submitter, I want the current signing key tried before deprecated keys, so that the active identity takes precedence.

17. As an Arkade flow submitter, I want deprecated keys tried in configured order, so that fallback behavior is deterministic.

18. As a maintainer, I want signer resolution extracted into a deep, unit-testable helper, so that all normal signing paths share the same key-matching behavior.

19. As a maintainer, I want finalization to remember which signer matched each signed input, so that follow-up signatures use the same historical key.

20. As a maintainer, I want environment key parsing isolated from application signing, so that application signing behavior stays independent from deployment configuration mechanics.

21. As a maintainer, I want generated API code updated with the schema change, so that checked-in generated sources remain consistent.

22. As a maintainer, I want README documentation updated, so that operators and clients understand current and deprecated signing keys.

23. As a maintainer, I want integration coverage for every signing endpoint using deprecated keys, so that key rotation behavior is protected end to end.

24. As a maintainer, I want onchain deprecated-key signing verified locally, so that the test proves the returned PSBT contains a valid deprecated-key signature.

25. As a maintainer, I want regtest configured with at least two deprecated keys, so that tests can prove behavior with non-first deprecated keys.

#### Configuration

Keep environment variables as the only signing-key configuration interface. Secret managers, parameter stores, and orchestration platforms should provide key material to introspector by injecting environment variables.

Keep the current signing key:

- External env var: `INTROSPECTOR_SECRET_KEY`
- Represents the current signing key

Add deprecated signing keys:

- External env var: `INTROSPECTOR_DEPRECATED_KEYS`
- Format: strict comma-separated private keys
- Unset or empty means no deprecated keys
- Reject spaces
- Reject empty entries
- Reject leading commas
- Reject trailing commas
- Reject double commas
- Each key must be valid hex
- Each decoded key must be exactly 32 bytes
- After length validation, use the existing private-key construction check
- Reject duplicate raw private-key bytes, including the current key repeated in deprecated keys
- Preserve deprecated key order internally

Update config shape:

- Keep `INTROSPECTOR_SECRET_KEY` as the current signing key
- Internally rename the loaded key concept to current key
- Add deprecated keys to config
- Keep environment parsing above the application service

#### Application Service

Update service construction to accept:

- Current private key
- Deprecated private keys
- Ark server URL

The application constructor should:

- Reject nil current key
- Reject nil deprecated key entries
- Wrap all keys into the existing signer type
- Store signers, not raw private-key slices

The service should store:

- Current signer
- Deprecated signers

#### Signer Resolution For Normal Signing Paths

Extract shared signer resolution into an unexported helper near the existing signer implementation.

Use this generic resolver for:

- Offchain transaction signing
- Intent proof signing
- Onchain transaction signing

The resolver should:

- Accept current signer and deprecated signers separately
- Try the current signer first
- Then try deprecated signers in configured order
- Continue to the next signer only on signer-mismatch
- Return both the matched signer and parsed Arkade script
- Use first-match-wins behavior if multiple signers could match

Endpoint-specific error handling should remain unchanged:

- Offchain transaction signing keeps current strict behavior and existing unrelated-signer skip semantics
- Onchain transaction signing keeps current strict behavior and existing unrelated-signer skip semantics
- Intent proof signing keeps current lenient behavior for invalid or non-matching Arkade-script entries

Mixed-key requests are allowed for endpoints using the generic resolver: different inputs in one request may resolve to different configured introspector keys.

#### Finalization Signing With Deprecated Keys

`SubmitFinalization` must support deprecated signing keys, but it should not use the generic per-entry signer resolver used by offchain transaction signing, intent proof signing, and onchain transaction signing.

Finalization has a different flow:

1. Inspect the signed intent proof.
2. Find Arkade script inputs already signed by this introspector.
3. Determine which configured signer produced or validates each proof signature.
4. Carry that signer association forward.
5. Use the same signer when signing matching forfeit or commitment inputs.

The finalization flow should build an internal association containing:

- Spent outpoint
- Parsed Arkade script
- Matched signer

This association is derived from the signed intent proof, not from the forfeit or commitment transaction itself.

Finalization behavior:

- Try the current signer first when inspecting proof signatures
- Then try deprecated signers in configured order
- A proof input is considered owned by a signer only if the Arkade script can be read for that signer’s public key
- A proof input is considered owned by a signer only if the proof contains a tapscript signature for the corresponding tweaked Arkade script key
- A proof input is considered owned by a signer only if that signature verifies against the script key
- If an input is matched to a signer, store that signer with the signed input
- When a forfeit spends a matched outpoint, sign the forfeit input with the stored signer
- When the commitment transaction spends a remaining matched outpoint, sign the commitment input with the stored signer
- Do not re-resolve signer ownership from the forfeit or commitment transaction
- Do not sign finalization inputs with the current signer unless the current signer was the one matched in the signed intent proof
- Inputs matched by different configured signers may coexist in one finalization request

Finalization should preserve existing validation behavior:

- The signed intent proof must still contain at least one input signed by a configured introspector signer
- Malformed forfeits remain rejected
- Connector-tree validation remains unchanged
- Commitment signing behavior remains unchanged except for selecting the matched signer
- Signed inputs are still removed once matched to a forfeit, leaving only unmatched boarding inputs for commitment signing

#### Public API

Extend `GetInfo` response:

- Keep `signer_pubkey` as the current signer public key
- Add `deprecated_signer_pubkeys`
- Deprecated signer public keys use the same compressed hex encoding as the current signer public key
- No deprecated keys means an empty list

Update:

- Protobuf schema
- Generated protobuf Go code
- Application info model
- gRPC handler mapping
- Public Go client info model
- Public Go client response mapping
- README API docs

#### Regtest Configuration

Update the regtest Docker Compose environment:

- Add at least two deterministic, test-only deprecated private keys
- Tests should use the last deprecated public key returned by `GetInfo`

#### Unit Tests: Key Parsing

Add config key parsing tests covering:

1. Valid current key parses successfully.

2. Empty deprecated keys parses as zero deprecated keys.

3. A single deprecated key parses successfully.

4. Multiple deprecated keys parse successfully.

5. Deprecated key order is preserved internally.

6. Invalid current key hex is rejected.

7. Invalid deprecated key hex is rejected.

8. Current key with decoded length other than 32 bytes is rejected.

9. Deprecated key with decoded length other than 32 bytes is rejected.

10. Deprecated keys with leading comma are rejected.

11. Deprecated keys with trailing comma are rejected.

12. Deprecated keys with double comma are rejected.

13. Deprecated keys with whitespace are rejected.

14. Current key duplicated in deprecated keys is rejected.

15. Deprecated key duplicated in deprecated keys is rejected.

#### Unit Tests: Signer Resolution

Add application-level signer resolution tests covering:

1. Current signer matches an Arkade script entry.

2. Deprecated signer matches an Arkade script entry.

3. Current signer is preferred when it matches.

4. Deprecated signers are tried after the current signer.

5. The last deprecated signer can match.

6. Signer fallback only continues on signer-mismatch.

7. Structural Arkade script errors are returned instead of hidden by fallback.

8. Resolver returns both the matched signer and parsed Arkade script.

9. Mixed-key inputs can resolve independently.

10. First matching signer wins if multiple configured signers could match.

#### Unit Tests: Finalization Signer Association

Add focused finalization tests covering:

1. Current-key proof signature is detected and associated with the current signer.

2. Deprecated-key proof signature is detected and associated with the matching deprecated signer.

3. Finalization signs a matching forfeit with the signer associated from the intent proof.

4. Finalization signs a matching commitment input with the signer associated from the intent proof.

5. Finalization does not sign with the current signer when the proof signature was produced by a deprecated signer.

6. A finalization request can contain proof inputs matched to different configured signers.

7. Malformed proof/script/signature cases preserve existing finalization failure behavior.

8. Inputs not signed by any configured signer are ignored or rejected according to the existing finalization rules.

#### Integration Test: Deprecated Key Signing

Add `TestSigningWithDeprecatedKey`.

Use at least two deprecated keys in the regtest environment.

The test should:

1. Call `GetInfo`.

2. Read `deprecated_signer_pubkeys`.

3. Select the last deprecated public key.

4. Build scripts using that deprecated public key.

5. Run subtests for every signing endpoint.

Subtests:

1. `SubmitTx` signs an offchain transaction using the deprecated key and completes the existing happy-path flow.

2. `SubmitIntent` signs an intent proof using the deprecated key and completes the flow through arkd where applicable.

3. `SubmitFinalization` completes a flow where the intent proof was signed using the last deprecated key, then verifies finalization succeeds using the same deprecated-key association.

4. `SubmitOnchainTx` signs an onchain PSBT using the deprecated key and locally verifies the returned PSBT signature.

Prefer adapting existing happy-path integration fixtures rather than creating unrelated new scenarios.

### Documentation

Update README configuration docs for:

- `INTROSPECTOR_DEPRECATED_KEYS`

Update README `GetInfo` docs for:

- Current signer public key
- Deprecated signer public keys

EDIT: removed the `INTROSPECTOR_KEY_SOURCE` switch requirement as it is simpler to handle encrypted keys at the deployment platform level, which should be able to inject secrets as environment variables.

chris-ricketts:
Closes #83 

**Summary**
- Supports `INTROSPECTOR_DEPRECATED_KEYS` as a strict comma-separated list of deprecated private keys, with validation for malformed, duplicate, zero, out-of-range, or current-key entries.
- Resolves Arkade script ownership against the current signer first, then deprecated signers, and signs with the matched key across `SubmitTx`, `SubmitIntent`, `SubmitFinalization`, and `SubmitOnchainTx`.
- Exposes deprecated signer public keys via `GetInfo` in protobuf, OpenAPI, gRPC handler, and client transport types.
- Adds unit coverage for config parsing and signer resolution, plus regtest coverage for deprecated-key signing across offchain, intent/finalization, and onchain flows.
- Updates README examples for camelCase JSON fields and documents the new deprecated key env var.
- Fixes the buf Docker image by installing pinned tool versions via `go install` and removing the hardcoded Linux armv7 binary.

**Testing**
- Added `internal/config/config_test.go`
- Added `internal/application/signer_test.go`
- Added `test/deprecated_key_signing_test.go`


ghost (commented):
## 🔍 Arkana Code Review — `feat: support both current and deprecated signing keys`

**Verdict: Code looks correct. Requesting human sign-off — this is protocol-critical (touches all signing paths).**

---

### Architecture

The approach is sound: resolve which signer (current or deprecated) owns each input, then sign with the matched key. The resolution logic correctly prioritizes the current key, falls back through deprecated keys in order, and short-circuits on structural errors while continuing on key-not-found errors.

The change is **backward compatible**. The protobuf field (`deprecated_signer_pubkeys = 3`) is additive. The `application.New()` signature change is internal-only. Downstream consumers (`bancod`, `layerzero-usdt0-arkade-demo`) will continue to work without changes — they just won't see the new `DeprecatedSignerPublicKeys` field until they bump their go.mod dependency.

---

### ⚠️ Protocol-Critical: Signing Path Changes

All four signing endpoints are modified:
- **SubmitTx** (`tx.go`): Uses `resolveArkadeScriptSigner` → signs with matched key
- **SubmitIntent** (`intent.go`): Same pattern, including the message input (index 0) signing
- **SubmitFinalization** (`finalization.go`): Uses `getSignedInputAssociations` for signature verification + signer pairing on forfeits and commitment tx
- **SubmitOnchainTx** (`onchain.go`): Same pattern as SubmitTx

**I've verified:**
- ✅ Current key always gets priority (tried first in all paths)
- ✅ Each input independently resolves its signer — a tx can mix current and deprecated key inputs
- ✅ Structural errors from `ReadArkadeScript` (malformed PSBT, unsupported closure type, etc.) propagate immediately and correctly — they're key-independent, so no point trying other candidates
- ✅ `ErrTweakedArkadePubKeyNotFound` correctly triggers fallback to next candidate
- ✅ In `getSignedInputAssociations` (`finalization.go:163-197`): if `ReadArkadeScript` succeeds for a candidate but sig verification fails, it returns an error immediately rather than trying other candidates. This is correct — `ReadArkadeScript` matches on the tweaked pubkey (which is unique per base key + script hash), so if the tweaked key is found but the signature is invalid, the PSBT is genuinely malformed, not a key mismatch.
- ✅ Checkpoint signing in `tx.go:89` correctly uses the same `matchedSigner` as the corresponding ark tx input

---

### 🔐 Security

- ✅ **Key validation** (`config.go:100-121`): `parsePrivateKey` is a significant improvement over the original. It validates: hex encoding, exact 32-byte length, non-zero scalar, and scalar < curve order (via `ModNScalar.SetBytes` overflow check). This prevents `PrivKeyFromBytes` from silently reducing keys ≥ N.
- ✅ **Duplicate detection** (`config.go:68-82`): Seeds dedup map with current key before parsing deprecated keys. Catches both current-in-deprecated and deprecated-in-deprecated duplicates.
- ✅ **Strict CSV parsing**: No trimming, no empty entries — leading/trailing commas, double commas, whitespace all rejected. Good.
- ✅ **Defensive copies** in `GetInfo` (`service.go:122`, `introspector_handler.go:37`): `append([]string(nil), ...)` prevents callers from mutating internal state.
- ✅ **Public keys only exposed via API**: `DeprecatedSignerPubkeys` exposes compressed public keys, not private keys.

---

### 📋 Minor Notes

1. **`finalization.go:163`** — `candidates := append([]signer{current}, deprecated...)` allocates a new slice per input in the outer loop. Negligible for typical workloads (few inputs × few deprecated keys), but could be hoisted above the loop if this ever becomes a hot path.

2. **`signer.go:83`** — When `deprecated` is empty and current doesn't match, `resolveArkadeScriptSigner` returns `err` from the current-signer attempt. This works correctly (returns `ErrTweakedArkadePubKeyNotFound`), but the variable scoping makes it slightly non-obvious. A named error variable would improve readability. Very minor.

3. **`intent.go:48`** — The caller continues on ANY error from `resolveArkadeScriptSigner`, which matches the original behavior (continue on any `ReadArkadeScript` error). This is intentional: unresolvable entries are skipped since they may belong to a different signer entirely.

---

### ✅ Test Coverage

- `config_test.go`: 16 test cases covering valid configs, edge cases (zero key, curve order, above curve order), strict CSV validation, and duplicate detection. Thorough.
- `signer_test.go`: Tests current match, deprecated match, current-key preference, ordered deprecated fallback, mixed-key entries, structural errors, and exhausted-signers errors. Thorough.
- `deprecated_key_signing_test.go`: Integration tests covering all three signing flows (offchain, intent+finalization, onchain) with deprecated keys. Good end-to-end coverage.

---

### Cross-Repo Impact

| Consumer | Impact | Action Needed |
|----------|--------|---------------|
| `bancod` | None — additive proto field, internal API changes only | Bump go.mod when ready to consume deprecated pubkeys |
| `layerzero-usdt0-arkade-demo` | None — uses `replace` directives | Will pick up changes on next local introspector update |
| `introspector-enclave` | None — doesn't import introspector packages | N/A |

No companion PRs needed.

---

### 🏗️ Buf Dockerfile

The fix (`buf.Dockerfile`) is good — replaces a hardcoded `Linux-armv7` binary with `go install` at pinned versions. This makes the build portable across architectures. Minor but welcome improvement.

---

**🚨 This PR is protocol-critical (modifies signing logic for VTXOs, forfeits, commitment transactions, and onchain exits). Human sign-off required before merge.**

ghost (commented):
## 🔍 Arkana — Re-review after force push

Diff is identical (md5 `dc6e4d1977091e426cecb20537256a60` matches pre- and post-push). This was a clean rebase — no code changes.

My [previous review](https://github.com/ArkLabsHQ/introspector/pull/86#pullrequestreview-) still applies in full:
- ✅ Protocol correctness verified across all four signing paths
- ✅ Security (key validation, dedup, defensive copies) verified
- ✅ Test coverage is thorough
- ✅ No cross-repo breakage

**Still requesting human sign-off** — this is protocol-critical (touches all signing paths). No new findings.

louisinger (approved):

PR #88, opened 2026-05-22
Update sighash template for introspector sigs and add OP_SIGHASH

msinkec:
## Summary

Adds `OP_SIGHASH` (`0xf6`) and migrates the arkade VM's tapscript signature-verification path (`OP_CHECKSIG`, `OP_CHECKSIGVERIFY`, `OP_CHECKSIGADD`, and the new `OP_SIGHASH`) onto a non-standard, domain-separated tapscript signature hash. Bitcoin-level PSBT signing in `internal/application/finalization.go` is untouched.

## OP_SIGHASH

| Hex | Stack | Tapscript-only |
|---|---|---|
| `0xf6` | `hashType → sighash32` | yes |

Pops a sighash flag: a minimally encoded scriptNum in `[0,255]`, which must be one of `{0x00, 0x01, 0x02, 0x03, 0x81, 0x82, 0x83}`. Pushes the 32-byte arkade tapscript signature hash for the currently executing input. The pushed digest equals the message `OP_CHECKSIG` verifies in the same execution context, so scripts can compute it, sign over it, and verify with `OP_CHECKSIGFROMSTACK`.

## Sighash template — two deliberate departures from BIP342

1. **Witness-blob masking.** When `sha_outputs` and the per-output digest used by `SIGHASH_SINGLE` are computed, the single `OP_RETURN` carrying an introspector packet has every entry rewritten with `witness_len = 0` and the witness bytes dropped. Script bytes, `vin`, entry count, co-located ARK packets such as the asset packet, and every non-extension output continue to flow into the digest unchanged.

   This lets scripts be pre-signed before any party has supplied runtime witness arguments, and lets witness args be re-supplied per spend attempt without invalidating signatures.

2. **BIP-340 tag `"ArkadeTapSighash"`**, not BIP342's `"TapSighash"`. The digest domains are disjoint by construction: a signature valid under one cannot pass verification under the other, even when the message bytes happen to match. Bitcoin-level signatures continue to use `"TapSighash"` via `txscript.CalcTapscriptSignaturehash`.

### Side effect: OP_CHECKSIG correctness

The previous `baseTapscriptSigVerifier.Verify` built a BIP341 keypath digest in tapscript context, so it would have rejected every real BIP342 signature. The new shared helper fixes that bug and is the path `OP_CHECKSIG`, `OP_CHECKSIGVERIFY`, `OP_CHECKSIGADD`, and `OP_SIGHASH` all go through, so the four cannot drift.

## Tests

- `TestArkadeSighashByteLayoutMatchesBIP342` — for every valid sighash flag, `TaggedHash("TapSighash", buildArkadeSigMsg(originalTx, flag))` equals `txscript.CalcTapscriptSignaturehash` over the witness-masked tx. Pins the byte layout to BIP342 byte-for-byte; the only departures from BIP342 are witness masking and tag.
- `TestArkadeSighashMasksWitnessBlobs` — witness-content and witness-length mutations do not change the digest; mutating entry script bytes or `vin`, the co-located asset packet, or any non-extension output's value does.
- `TestArkadeSighashIsDomainSeparated` — arkade digest ≠ BIP342 digest for a tx with no introspector packet, where the tag is the only difference.
- `TestOpSighashMatchesCheckSigFromStack` — end-to-end round trip through `OP_SIGHASH` and `OP_CHECKSIGFROMSTACK` on a real tapscript spend.
- `TestOpCheckSigArkadeSighash` — locks in the verifier fix; `OP_CHECKSIG` accepts a signature over the arkade digest.
- `sighashSpec` in `opcode_test.go` — covers all valid flags and rejection cases: no context, underflow, non-minimal, out-of-range, unknown flag, and `SIGHASH_SINGLE` with no matching output.
- `sighashCaseBuilder` added to `FuzzOpcodes`.


ghost (changes requested):
## Aggressive Code Review — PR #88: OP_SIGHASH + Non-standard Arkade Tapscript Sighash

**Reviewer**: Arkana (AI)  
**Verdict**: Request changes — one doc bug and one fuzz concern, plus this is **protocol-critical** (signature verification) and requires human sign-off before merge.

---

### Overall Assessment

This is a well-structured, well-tested PR that:
1. **Fixes a real bug**: `baseTapscriptSigVerifier.Verify()` previously called `txscript.CalcTaprootSignatureHash` (BIP341 **keypath**), not `CalcTapscriptSignaturehash` (BIP342 tapscript). Every real tapscript signature would have been rejected.
2. **Introduces witness masking** so scripts can be pre-signed before runtime witness arguments are supplied.
3. **Adds domain separation** via `"ArkadeTapSighash"` tag, preventing cross-domain signature replay.
4. **Ships with comprehensive tests**: byte-layout equivalence, masking properties, domain separation, round-trip signing, opcode spec coverage, fuzz builder.

The `buildArkadeSigMsg` implementation faithfully mirrors the BIP342 sigMsg structure (epoch, hash_type, nVersion, nLockTime, cross-input midstates, sha_outputs, spend_type, per-input data, annex, tapscript extension) — I verified each field against BIP342 §4.

Cross-repo impact: **None**. All changed APIs (`computeArkadeSighash`, `buildArkadeSigMsg`, `maskExtensionOutput`, `arkadeOutputsHash`, `serializeIntrospectorPacketMasked`) are unexported. Downstream consumers (bancod, layerzero-usdt0-arkade-demo) only use exported APIs (`NewPacket`, `ComputeArkadeScriptPublicKey`, `ArkadeScriptHash`, opcode constants) which are untouched.

---

### Issues

#### 1. 🔴 Doc bug in `opcodeSighash` — `opcode.go:2771`

The docstring says:

> pushes the 32-byte **BIP342 tapscript signature hash**

But the opcode pushes the **arkade** (non-standard) digest, tagged `"ArkadeTapSighash"`. This contradicts the README, the PR description, and the implementation. Should read something like:

```
// opcodeSighash pops a sighash flag from the stack and pushes the 32-byte
// non-standard arkade tapscript signature hash of the currently executing
// input under that flag.
```

#### 2. 🟡 Fuzz panic on short data — `opcode_fuzz_test.go:196-209`

`sighashCaseBuilder.Build()` accesses `data[0]` and `data[1]` unconditionally. If the fuzzer ever supplies `data` shorter than 2 bytes, this panics. Other case builders (e.g. `taprootCheckSigCaseBuilder`) appear to have similar patterns, so this may be a pre-existing convention, but worth adding a guard:

```go
if len(data) < 2 {
    return defaultCaseBuilder{}.Build(data, world)
}
```

#### 3. 🟢 `maskExtensionOutput` fail-open behavior — `sigvalidate.go` (maskExtensionOutput)

The function returns `(nil, -1, nil)` when parsing fails at any boundary (extension parse error, type assertion failure, deserialization failure). This is commented as intentional ("masking is fail-open at any parsing boundary so a corrupted OP_RETURN cannot disable digest computation").

I've verified this is **fail-safe from a security perspective**: skipping the mask means the digest commits to MORE data (the full witness), not less. Signatures become more restrictive, not less. A pre-signed signature simply won't validate if the extension is malformed — which is the safe failure mode.

**No change needed**, but consider adding an explicit comment at the function level stating "fail-open is fail-safe because unmasked = commits to more data."

#### 4. 🟢 Error swallowing in `baseTapscriptSigVerifier.Verify()` — `sigvalidate.go:503`

When `computeArkadeSighash` returns an error, `Verify()` returns `verifyResult{}` (sigValid=false) — silently treating the error as "invalid signature". This was inherited from the old code (which had a `TODO(roasbeef): propagate the error here?`). The PR removes the TODO but keeps the behavior.

This is acceptable for now — a digest computation error in a well-formed engine state would indicate a programming bug, and treating it as "invalid" is safe. But a structured log or debug hook here would help diagnose issues in production.

---

### Correctness Verification

| Check | Result |
|---|---|
| BIP342 sigMsg field ordering (epoch → hash_type → nVersion → nLockTime → midstates → sha_outputs → spend_type → per-input → annex → tapscript ext) | ✅ Matches |
| ANYONECANPAY drops cross-input midstates, adds per-input outpoint+amount+sequence | ✅ Correct |
| SIGHASH_NONE drops sha_outputs, no per-output | ✅ Correct |
| SIGHASH_SINGLE drops sha_outputs, adds SHA256(output[idx]) with masking | ✅ Correct |
| SIGHASH_DEFAULT treated as SIGHASH_ALL for outputs | ✅ Correct (`0x00 & 0x03 == 0`, which is neither SINGLE nor NONE) |
| ext_flag=1 for tapscript, spend_type = 2*1 + annex_present | ✅ Correct |
| Tapscript extension: leafHash + key_version(0x00) + codeSepPos | ✅ Correct |
| Domain separation: `"ArkadeTapSighash"` ≠ `"TapSighash"` | ✅ Verified |
| Witness masking: witness_len=0 in masked serialization | ✅ Correct |
| `serializeIntrospectorPacketMasked` preserves entry count, vin, script — only drops witness | ✅ Verified against `Serialize()` |
| `isValidTaprootSigHash` accepts exactly {0x00,0x01,0x02,0x03,0x81,0x82,0x83} | ✅ Matches BIP341 |
| `opcodeSighash` validates flag range, hash type, and SIGHASH_SINGLE output existence before computing | ✅ Correct |
| `OP_CHECKSIG`, `OP_CHECKSIGVERIFY`, `OP_CHECKSIGADD` all route through `baseTapscriptSigVerifier.Verify()` → `computeArkadeSighash` | ✅ Verified call chain |
| `taprootSigVerifier.Verify()` (keypath) still uses BIP341 `CalcTaprootSignatureHash` | ✅ Unchanged, correct |
| `engine.go` now stores `tapLeaf` alongside `tapLeafHash` | ✅ Both populated in `verifyWitnessProgram` |

---

### ⚠️ PROTOCOL-CRITICAL FLAG

This PR changes **how every signature is verified inside the arkade VM** — `OP_CHECKSIG`, `OP_CHECKSIGVERIFY`, and `OP_CHECKSIGADD` all now use the non-standard arkade sighash instead of the (buggy) BIP341 keypath hash. This means:

- All existing arkade scripts that use signature verification will now verify against a different digest
- Any pre-existing signed transactions in the wild would need to be re-signed against the new digest

**This requires explicit human sign-off.** The code looks correct to me, and the test suite is thorough, but signature verification changes in a Bitcoin L2 demand a second pair of human eyes.

Fix the docstring (issue #1), consider the fuzz guard (issue #2), and get a human to approve.

ghost (approved):
## Follow-up Review — New Commits Since May 22

**Reviewer**: Arkana (AI)
**Verdict**: Approve — previous issues addressed, new changes are solid. **Still protocol-critical** — human sign-off required before merge.

---

### Previous Issues Status

| # | Issue | Status |
|---|-------|--------|
| 1 | 🔴 Doc bug `opcodeSighash` said "BIP342" | ✅ **Fixed** — now says "non-standard arkade tapscript signature hash" (`opcode.go:2793`) |
| 2 | 🟡 Fuzz panic on short data | ✅ **Fixed** — `if len(data) < 2 { return c }` guard added (`opcode_fuzz_test.go:197`) |
| 3 | 🟢 `maskExtensionOutput` fail-open comment | Not addressed — still a nice-to-have, not blocking |
| 4 | 🟢 Error swallowing in `Verify()` | Acknowledged — TODO removed, behavior unchanged (fail-closed is safe) |

---

### New Changes Analysis (10 commits since last review)

#### 1. ✅ Keypath verification removed (`sigvalidate.go`)

The old `taprootSigVerifier` (BIP341 keypath) and `baseTapscriptSigVerifier` (two-layer abstraction) have been collapsed into a single `tapscriptSigVerifier`. This is correct — arkade is tapscript-only, the keypath verifier was dead code, and its removal shrinks the attack surface. The `signatureVerifier` interface and `verifyResult` struct are also gone. Clean simplification.

The critical fix is preserved: `Verify()` now calls `computeArkadeSighash` instead of `txscript.CalcTaprootSignatureHash` (the old BIP341 keypath bug).

#### 2. ✅ `trackCodeSep` flag (`engine.go:53`, `script.go:181`)

New boolean prevents OP_CODESEPARATOR in arkade scripts from updating `codeSepPos`. Rationale: the sighash commits to the *spending tapleaf* hash and its code separator position, but arkade scripts execute from the introspector packet, not the spending tapleaf. If an arkade script contains OP_CODESEPARATOR, it must not shift the digest away from what was pre-signed.

- Default is `true` (standard tapscript path in `verifyWitnessProgram`)
- Explicitly set to `false` in `ArkadeScript.Execute()`
- Well-tested by `TestArkadeScriptExecuteDoesNotUsePacketCodeSeparatorForSighash`

#### 3. ✅ `CalcTapscriptSignaturehash` public API (`sigvalidate.go`)

New exported function for external signers to compute the arkade sighash without running the engine. Input validation is correct (nil checks for sigHashes, tx, prevOutFetcher). Creates a minimal `Engine` internally.

**One gap** (non-blocking): this API doesn't accept an annex or codeSepPos. It always uses `annex=nil` and `codeSepPos=0xFFFFFFFF`. This is correct for the current Ark protocol (no annexes, arkade scripts don't track codeSep), but if annex support is ever needed, the API will need extending. Consider documenting this limitation in the function's godoc.

#### 4. ✅ `spendingTapLeaf` rename + leaf version check (`script.go`)

`tapLeaf` → `spendingTapLeaf` clarifies that this is the Bitcoin spending tapleaf, not the arkade script. The new `LeafVersion != BaseLeafVersion` check in `ReadArkadeScript` (`script.go:76-78`) rejects unknown leaf versions early — good defensive measure.

#### 5. ✅ `newTaprootExecutionCtxForLeaf` constructor (`engine.go:86-93`)

Eliminates the scattered pattern of creating a context then manually setting tapLeaf + tapLeafHash. Used consistently in both `verifyWitnessProgram` and `ArkadeScript.Execute`. Reduces chance of forgetting to set one of the two fields.

#### 6. ✅ New tests

Substantial test additions covering the new functionality:
- `TestOpSighashMatchesCheckSigFromStack` — round-trip: OP_SIGHASH → sign → OP_CHECKSIGFROMSTACK
- `TestOpCheckSigArkadeSighash` — OP_CHECKSIG verifies arkade sighash (not BIP342)
- `TestArkadeSighashSingleMasksExtensionOutput` — SIGHASH_SINGLE per-output masking when idx hits extension
- `TestArkadeSighashByteLayoutMatchesBIP342WithAnnexAndCodeSep` — annex + codeSep byte-layout match
- `TestArkadeScriptExecuteUsesSpendingTapLeafForSighash` — sighash commits to spending tapleaf, not arkade script
- `TestArkadeScriptExecuteDoesNotUsePacketCodeSeparatorForSighash` — OP_CODESEPARATOR in arkade script doesn't shift digest
- `TestReadArkadeScriptRejectsNonBaseSpendingTapLeafVersion` — leaf version guard
- `TestSignedPayToOutput` — integration test: signed-pay-to-output pattern end-to-end
- `sighashSpec()` — opcode spec with valid/invalid vectors

#### 7. ✅ Cross-repo impact

All removed types (`signatureVerifier`, `verifyResult`, `taprootSigVerifier`, `baseTapscriptSigVerifier`) were unexported. Downstream consumers (bancod, layerzero-usdt0-arkade-demo) only use exported APIs which are untouched. The new `CalcTapscriptSignaturehash` export adds API surface but breaks nothing. **No cross-repo breakage.**

---

### Minor Observations (non-blocking)

1. **`replaceIntrospectorPacket` test helper** (`signed_pay_to_output_test.go:237-259`): Creates a new extension with *only* the introspector packet, dropping any pre-existing asset packet. Fine for this test (no asset packet needed), but the helper name suggests it replaces just the introspector packet within an existing extension. A comment clarifying this would prevent misuse in future tests.

2. **`CalcTapscriptSignaturehash` annex gap**: As noted above, document that this API assumes no annex and blank codeSepPos. This matches all current usage patterns but should be explicit.

---

### Verdict

The PR is in good shape. Both previous issues are fixed. The new changes (keypath removal, trackCodeSep, public sighash API, leaf version check) are correct and well-tested. The byte-layout equivalence tests against btcd's BIP342 implementation provide strong confidence that the only departures from BIP342 are the intentional witness masking and domain tag.

**⚠️ Protocol-critical**: This PR changes signature verification and sighash computation. Human sign-off required before merge.

louisinger (approved):

pkg/arkade/sigvalidate.go
```diff
@@ -1,30 +1,320 @@
...
+	return digest[:], nil
+}
+
+// CalcTapscriptSignaturehash returns the non-standard arkade tapscript
+// signature hash used by OP_CHECKSIG and OP_SIGHASH inside arkade scripts.
+//
+// The byte layout mirrors BIP342's tapscript sigMsg with arkade's witness
+// masking and "ArkadeTapSighash" final tag. Callers should pass the active
+// Bitcoin spending tapleaf whose hash the signature commits to.
+func CalcTapscriptSignaturehash(
```
chris-ricketts, line 82: Nit: lower case "hash" looks weird, would expect `CalcTapscriptSignatureHash` to match [bctd](https://github.com/btcsuite/btcd/blob/master/txscript/sighash.go#L611)

As this is a public API, it might be worth changing now

pkg/arkade/sigvalidate.go
```diff
@@ -1,30 +1,320 @@
...
+	if !anyoneCanPay {
+		sigMsg.Write(hashCache.HashPrevOutsV1[:])
+		sigMsg.Write(hashCache.HashInputAmountsV1[:])
+		sigMsg.Write(hashCache.HashInputScriptsV1[:])
+		sigMsg.Write(hashCache.HashSequenceV1[:])
+	}
+
+	// 5. sha_outputs, unless SIGHASH_SINGLE or SIGHASH_NONE drop it. The
+	// SINGLE-specific per-output digest goes in further below.
+	sigHashMode := hashType & 0x03
```
chris-ricketts: Nit: use canonical Sighash Mask: `0x1f` - `0x03` is valid due to current `isValidTaprootSigHash` implementation

chris-ricketts (approved):

pkg/arkade/sigvalidate.go
chris-ricketts: Ignore, I was looking at the wrong function - it matches [btcd](https://github.com/btcsuite/btcd/blob/master/txscript/sighash.go#L628)

pkg/arkade/sigvalidate.go
chris-ricketts: added in [260f72e](https://github.com/ArkLabsHQ/introspector/pull/88/commits/260f72e0d5e3cd62c36040a5c9ff8e8a1f17e645)

Issue #81, opened 2026-05-19
Unify compute-budget system for heavy opcodes


msinkec:
The Arkade VM enforces script size (`MaxScriptSize = 10000`), stack size (`MaxStackSize = 1000`), and BIP-342's per-input sigops budget — but it does **not** enforce a per-script op-count limit. BIP-342 dropped that limit on purpose.

That trade-off is sound for plain Bitcoin tapscript, where the per-call cost of every opcode is small or already charged against the sigops budget.

Several heavy opcodes added to Arkade do not yet fit that model. They have per-call costs orders of magnitude above other opcodes, and the engine charges nothing against the sigops budget for them. They are only constrained by 10 KB script-size cap.

We should pick a design for a compute-budget brake.

## What the Engine Charges Today

Only the BIP-342 sigops budget exists, defined at `pkg/arkade/engine.go:48-77`:

```go
const sigOpsDelta = 50

func newTaprootExecutionCtx(inputWitnessSize int32) *taprootExecutionCtx {
    return &taprootExecutionCtx{
        sigOpsBudget: sigOpsDelta + inputWitnessSize,
    }
}
```

Starting budget per input is `50 + witness_size_bytes`; each charged signature opcode deducts 50.

Charged opcodes today, via callers of `tallysigOp()`:

- `OP_CHECKSIG` / `OP_CHECKSIGVERIFY` — `pkg/arkade/opcode.go:1782`
- `OP_CHECKSIGADD` — `pkg/arkade/opcode.go:1869`

Notably **not charged**, despite being non-trivial:

- `OP_CHECKSIGFROMSTACK` (~84 µs / call)
- `OP_ECMULSCALARVERIFY`
- `OP_TWEAKVERIFY`
- `OP_ECADD`
- `OP_ECMUL`
- `OP_ECPAIRING`
- `OP_MODEXP`

https://github.com/ArkLabsHQ/introspector/pull/79
https://github.com/ArkLabsHQ/introspector/pull/80

## Measured Per-Call Costs

Apple Silicon, single-threaded:

| Operation                                          | Per-call         | vs Schnorr verify |
| -------------------------------------------------- | ---------------: | ----------------: |
| `OP_ECADD` (any curve)                             | 1.5 – 3.7 µs     | ~0.03×            |
| `OP_ECMUL` (256-bit scalar, any curve)             | 25 – 45 µs       | 0.3 – 0.5×        |
[...]

PR #80, opened 2026-05-19
Implement OP_MODEXP

msinkec:
Adds `OP_MODEXP` (`0xd9`) for modular exponentiation, unlocking RSA/Diffie-Hellman-style constructions and Schnorr/ECDSA modular inverses in Arkade Script.

- Pops `base`, `exp`, `modulus` and pushes `base^exp mod modulus` as a canonical `BigNum` in `[0, modulus)`.
- Rejects `modulus <= 0` (new `ErrBigNumModulusNotPositive`) and `exp < 0` (new `ErrBigNumNegativeExponent`). Negative `base` is reduced canonically before exponentiation.
- Modular inverse is expressible without a dedicated opcode: for a prime modulus `p`, compute `a^(p-2) mod p` (Fermat's little theorem). A separate `OP_MODINV` for non-prime moduli is intentionally deferred.
- Backed by `math/big.Int.Exp` via a new `BigNum.Modexp` method.
- Renames the placeholder `OP_UNKNOWN217` to `OP_MODEXP`.

Adds `BigNum`-level and full opcode vector coverage (canonical residue, `exp=0`, `base=0`, `modulus=1`, negative-base canonicalization, Fermat inverse, RSA-sized cross-check, and rejection paths for non-positive modulus, negative exponent, non-minimal/oversized operands, and stack underflow). Updates the README opcode table.

PR #91, opened 2026-05-28
Opcode count compute limits

msinkec:
## Summary

Adds a per-input execution-count brake for the VM's heavy opcodes and removes the BIP-342 sigops budget, replacing both with one simple lookup table. This bounds worst-case single-input CPU and unblocks the heavy-opcode PRs (#79 EC ops, #80 OP_MODEXP), which add opcodes that are otherwise charged nothing and only limited by the 10 KB script-size cap.

## Motivation

The VM enforces script size and stack size but has no op-count limit. BIP-342 dropped it deliberately, which is fine for cheap opcodes.

Several Arkade opcodes (`OP_CHECKSIGFROMSTACK`, `OP_ECADD`, `OP_ECMUL`, `OP_ECPAIRING`, `OP_ECMULSCALARVERIFY`, `OP_TWEAKVERIFY`, `OP_MODEXP`) cost orders of magnitude more per call, so an adversarial 10 KB script could drive multi-second CPU per input. See #81.

## What changed

- Added `ComputeLimits` lookup table (`pkg/arkade/compute_limits.go`):
  - `type ComputeLimits map[byte]int`
  - Opcode → max executions per input
  - Absent opcode ⇒ unlimited
  - `DefaultComputeLimits()` returns a fresh copy

- Added engine charging (`engine.go`):
  - One `chargeOpcode` call in `executeOpcode`
  - Runs after the dead-branch return, so opcodes in unexecuted branches don't count
  - Runs before the handler
  - Per-input counts live on the tapscript context
  - Exhaustion returns `txscript.ErrScriptTooBig`, naming the opcode and limit

- Removed the BIP-342 sigops budget:
  - Deleted `sigOpsBudget`, `tallysigOp`, `sigOpsDelta`
  - Deleted witness-size scaling
  - Deleted both `tallysigOp` calls in the sig handlers
  - That budget models L1 block-weight/fee economics, which don't apply to the emulator
  - Sig opcodes are now entries in the same count-limit table

- Added admin override (`internal/config`):
  - Example: `EMULATOR_COMPUTE_LIMITS="OP_ECPAIRING=8,OP_MODEXP=128"`
  - Applied on top of defaults
  - Threaded via `WithComputeLimits` into the three `script.Execute` call sites
  - Unknown opcode, non-integer limit, or negative limit fails config load

- Retained existing per-call caps:
  - `maxECPairingCount = 16`
  - `maxModexpOperandLen = 64`
  - These bound a single call's cost
  - The count limit bounds the number of calls
  - They compose

## Default limits

| Opcode(s)                                                              | Limit   | ~cost/call |
| ---------------------------------------------------------------------- | ------- | ---------- |
| `OP_CHECKSIG`, `OP_CHECKSIGVERIFY`, `OP_CHECKSIGADD`, `OP_CHECKSIGFROMSTACK` | 50 each | 84 µs      |
| `OP_ECADD`                                                             | 1000    | 3.7 µs     |
| `OP_ECMUL`, `OP_ECMULSCALARVERIFY`, `OP_TWEAKVERIFY`                   | 50 each | 84 µs      |
| `OP_ECPAIRING`                                                         | 2       | 2.04 ms    |
| `OP_MODEXP`                                                            | 64      | 60 µs      |

## Behavior change

Dropping the sigops budget removes the BIP-342 empty-signature exemption: an empty-signature `OP_CHECKSIG` / `OP_CHECKSIGADD` in the unselected-branch multisig pattern now counts toward the opcode's limit.

This only over-counts cheap non-verifying calls. It never permits extra real verifications, and 50 is generous for realistic multisig.

---

Closing #81

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -125,7 +125,8 @@ func parseComputeLimits(raw string) (arkade.ComputeLimits, error) {
 	for pair := range strings.SplitSeq(raw, ",") {
 		pair = strings.TrimSpace(pair)
 		if pair == "" {
-			continue
+			return nil, fmt.Errorf(
+				"invalid empty compute limit override in %q", raw)
 		}
 		name, value, ok := strings.Cut(pair, "=")
 		if !ok {
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"strconv"
 	"strings"
 
 	arklib "github.com/arkade-os/arkd/pkg/ark-lib"
 	"github.com/arkade-os/emulator/internal/application"
+	"github.com/arkade-os/emulator/pkg/arkade"
 	"github.com/btcsuite/btcd/btcec/v2"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
@@ -23,6 +25,7 @@ const (
 	TLSExtraDomains = "TLS_EXTRA_DOMAINS"
 	LogLevel        = "LOG_LEVEL"
 	ArkdURL         = "ARKD_URL"
+	ComputeLimits   = "COMPUTE_LIMITS"
 )
 
 var (
@@ -43,6 +46,7 @@ type Config struct {
 	TLSExtraIPs     []string
 	TLSExtraDomains []string
 	ArkdURL         string
+	ComputeLimits   arkade.ComputeLimits
 }
 
 func LoadConfig() (*Config, error) {
@@ -84,6 +88,11 @@ func LoadConfig() (*Config, error) {
 	logLevel := viper.GetInt(LogLevel)
 	log.SetLevel(log.Level(logLevel))
 
+	computeLimits, err := parseComputeLimits(viper.GetString(ComputeLimits))
+	if err != nil {
+		return nil, err
+	}
+
 	cfg := &Config{
 		CurrentKey:      currentKey,
 		DeprecatedKeys:  deprecatedKeys,
@@ -93,11 +102,52 @@ func LoadConfig() (*Config, error) {
 		TLSExtraIPs:     viper.GetStringSlice(TLSExtraIPs),
 		TLSExtraDomains: viper.GetStringSlice(TLSExtraDomains),
 		ArkdURL:         viper.GetString(ArkdURL),
+		ComputeLimits:   computeLimits,
 	}
 	if cfg.ArkdURL == "" {
 		return nil, fmt.Errorf("missing arkd url")
 	}
 	return cfg, nil
+}
+
+// parseComputeLimits builds the per-input opcode compute brake from the
+// EMULATOR_COMPUTE_LIMITS override, applied on top of the engine defaults. The
+// override is a comma-separated list of OPCODE=limit pairs, e.g.
+// "OP_ECPAIRING=8,OP_MODEXP=128"; an empty value yields the defaults unchanged.
+// It errors on an unknown opcode name, a non-integer limit, or a negative
+// limit.
+func parseComputeLimits(raw string) (arkade.ComputeLimits, error) {
+	limits := arkade.DefaultComputeLimits()
+	if strings.TrimSpace(raw) == "" {
+		return limits, nil
+	}
+
+	for pair := range strings.SplitSeq(raw, ",") {
+		pair = strings.TrimSpace(pair)
+		if pair == "" {
+			continue
+		}
+		name, value, ok := strings.Cut(pair, "=")
+		if !ok {
+			return nil, fmt.Errorf(
+				"invalid compute limit override %q, want OPCODE=limit", pair)
+		}
+		name = strings.TrimSpace(name)
+		op, ok := arkade.OpcodeByName[name]
+		if !ok {
+			return nil, fmt.Errorf("unknown opcode %q in compute limits", name)
+		}
+		limit, err := strconv.Atoi(strings.TrimSpace(value))
+		if err != nil {
+			return nil, fmt.Errorf("invalid limit for opcode %q: %w", name, err)
+		}
+		limits[op] = limit
+	}
+
+	if err := limits.Validate(); err != nil {
+		return nil, err
+	}
+	return limits, nil
 }
 
 func parsePrivateKey(keyHex, name string) (*btcec.PrivateKey, error) {
@@ -124,5 +174,5 @@ func parsePrivateKey(keyHex, name string) (*btcec.PrivateKey, error) {
 }
 
 func (c *Config) AppService(ctx context.Context) (application.Service, error) {
-	return application.New(ctx, c.CurrentKey, c.DeprecatedKeys, c.ArkdURL)
+	return application.New(ctx, c.CurrentKey, c.DeprecatedKeys, c.ArkdURL, c.ComputeLimits)
 }
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -138,7 +138,12 @@ func parseComputeLimits(raw string) (arkade.ComputeLimits, error) {
 		if !ok {
 			return nil, fmt.Errorf("unknown opcode %q in compute limits", name)
 		}
-		limit, err := strconv.Atoi(strings.TrimSpace(value))
+		value = strings.TrimSpace(value)
+		if value == "" {
+			delete(limits, op)
+			continue
+		}
+		limit, err := strconv.Atoi(value)
 		if err != nil {
 			return nil, fmt.Errorf("invalid limit for opcode %q: %w", name, err)
 		}
--- a/pkg/arkade/compute_limits.go
+++ b/pkg/arkade/compute_limits.go
@@ -27,6 +27,11 @@ func (c ComputeLimits) Validate() error {
 // single-threaded) to bound worst-case per-input CPU well below the
 // multi-second cases an unbounded script could reach.
 func DefaultComputeLimits() ComputeLimits {
+	// Aggregate-cost note: this table is deliberately a simple per-opcode
+	// lookup, not a grouped or weighted budget. If every listed opcode is pushed
+	// to its independent cap, the measured heavy-opcode aggregate is ~37 ms per
+	// input on an Apple M4 Pro. That is looser than the ~24 ms grouped design,
+	// but keeps the policy easy to read, configure, and reason about for now.
 	return ComputeLimits{
 		// Schnorr-class signature verification, ~84 µs each.
 		OP_CHECKSIG:          50,
@@ -42,7 +47,3 @@ func DefaultComputeLimits() ComputeLimits {
 		OP_MODEXP:            64,   // ~60 µs at the 64-byte operand cap
 	}
 }
-
-// defaultComputeLimits is the shared, read-only default applied by NewEngine to
-// every engine that does not override it via WithComputeLimits.
-var defaultComputeLimits = DefaultComputeLimits()
--- a/pkg/arkade/compute_limits.go
+++ b/pkg/arkade/compute_limits.go
@@ -47,3 +47,14 @@ func DefaultComputeLimits() ComputeLimits {
 		OP_MODEXP:            64,   // ~60 µs at the 64-byte operand cap
 	}
 }
+
+func cloneComputeLimits(c ComputeLimits) ComputeLimits {
+	if c == nil {
+		return nil
+	}
+	clone := make(ComputeLimits, len(c))
+	for op, limit := range c {
+		clone[op] = limit
+	}
+	return clone
+}
--- a/pkg/arkade/compute_limits.go
+++ b/pkg/arkade/compute_limits.go
@@ -2,108 +2,47 @@ package arkade
 
 import "fmt"
 
-// OpcodeGroup is a named set of opcodes that share a single per-input
-// execution-count limit. Each time any member opcode executes, the group's
-// remaining count is decremented; execution fails once a group would go below
-// zero.
-type OpcodeGroup struct {
-	Name    string
-	Opcodes []byte
-	Limit   int
-}
-
-// ComputeLimits is the declarative compute brake: a set of opcode groups, each
-// with its own per-input execution-count cap. Opcodes that belong to no group
-// are unlimited, matching tapscript's lack of an op-count limit for the cheap
+// ComputeLimits maps an opcode to the maximum number of times it may execute
+// during a single input's script evaluation. Opcodes absent from the map are
+// unlimited, matching tapscript's lack of an op-count limit for the cheap
 // opcodes whose per-call cost is negligible.
-type ComputeLimits struct {
-	Groups []OpcodeGroup
-}
+type ComputeLimits map[byte]int
 
-// compiledComputeLimits is the hot-path form of a ComputeLimits: a constant-time
-// opcode→group lookup plus per-group metadata addressed by group index.
-type compiledComputeLimits struct {
-	groupOf [256]int // opcode byte -> group index, -1 if ungrouped
-	limits  []int    // per-group execution-count limit
-	names   []string // per-group name, for error messages
-}
-
-// Validate reports whether the configuration is well formed: every limit must
-// be non-negative and no opcode may appear in more than one group.
+// Validate reports whether every configured limit is non-negative.
 func (c ComputeLimits) Validate() error {
-	var seen [256]bool
-	for _, g := range c.Groups {
-		if g.Limit < 0 {
-			return fmt.Errorf("compute limits: group %q has negative limit %d",
-				g.Name, g.Limit)
-		}
-		for _, op := range g.Opcodes {
-			if seen[op] {
-				return fmt.Errorf("compute limits: opcode 0x%02x appears in "+
-					"more than one group", op)
-			}
-			seen[op] = true
+	for op, limit := range c {
+		if limit < 0 {
+			return fmt.Errorf("compute limit for %s is negative: %d",
+				opcodeArray[op].name, limit)
 		}
 	}
 	return nil
 }
 
-// compile builds the lookup form. It panics on an invalid configuration: a
-// malformed ComputeLimits is always a programmer error, never runtime input.
-// Use Validate to check a configuration without panicking.
-func (c ComputeLimits) compile() *compiledComputeLimits {
-	if err := c.Validate(); err != nil {
-		panic(err)
-	}
-	cl := &compiledComputeLimits{
-		limits: make([]int, len(c.Groups)),
-		names:  make([]string, len(c.Groups)),
-	}
-	for i := range cl.groupOf {
-		cl.groupOf[i] = -1
-	}
-	for gi, g := range c.Groups {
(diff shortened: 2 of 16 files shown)


ghost (changes requested):
## Code Review — Opcode Count Compute Limits

Good design. The flat per-opcode count table is simpler and more appropriate for Arkade than the BIP-342 witness-size-proportional sigops budget. The implementation is clean, well-tested, and the config-override path is solid. A few issues to address:

---

### 🔴 MEDIUM — Shared mutable map in `defaultComputeLimits` (`pkg/arkade/compute_limits.go:47`, `pkg/arkade/engine.go:835`)

`NewEngine` assigns the **shared** `defaultComputeLimits` map directly to `vm.limits`:

```go
var defaultComputeLimits = DefaultComputeLimits()  // one map for the process

// engine.go:835
limits: defaultComputeLimits,  // every engine gets the SAME map reference
```

`ComputeLimits` is `map[byte]int` — a reference type. Today no code writes through `vm.limits[op]`, so this is **safe**. But it's fragile: a future change that does `vm.limits[op] = X` on any engine would silently corrupt **every** engine's limits for the rest of the process lifetime.

**Fix:** Either:
- Copy the map in `NewEngine` (safe, tiny cost — 10 entries), or
- Keep the shared reference but add a prominent `// READ-ONLY — do not mutate` comment on both the var and the field, and add a test that verifies `defaultComputeLimits` is not mutated after engine creation.

Copying is the safer choice for protocol-critical code.

---

### 🟡 LOW — `WithComputeLimits` replaces the entire map (`pkg/arkade/script.go:40-43`)

```go
func WithComputeLimits(c ComputeLimits) ExecuteOption {
    return func(engine *Engine) {
        engine.limits = c
    }
}
```

If a caller passes a partial map (e.g., `ComputeLimits{OP_ECPAIRING: 8}`), **all other heavy opcodes become unlimited**. The config parser handles this correctly by starting from `DefaultComputeLimits()` and overriding, but a direct API caller of `WithComputeLimits` could easily make this mistake.

**Fix:** Add a godoc warning:

```go
// WithComputeLimits overrides the per-input opcode-execution compute brake for
// this execution. c replaces the default limits entirely — opcodes absent from
// c become unlimited. Callers that want to override a subset should start from
// DefaultComputeLimits() and modify the returned map.
```

---

### 🟡 LOW — `ErrScriptTooBig` reuse for semantically different error (`pkg/arkade/engine.go:340`)

The compute-limit exhaustion reuses `ErrScriptTooBig`:

```go
return scriptError(txscript.ErrScriptTooBig,
    fmt.Sprintf("opcode %s exceeded execution limit of %d", ...))
```

This is the same error code used at `engine.go:842` for "script size exceeds max allowed size". Callers that match on `ErrScriptTooBig` will conflate "script too large" with "opcode executed too many times." The descriptive string distinguishes them, but error-code consumers won't see the string.

Not blocking, but consider whether a dedicated `ErrComputeLimitExceeded` code would be cleaner for downstream error handling.

---

### 🟢 INFO — Observations (no action needed)

1. **Dead-branch skip is correct** — `chargeOpcode` is called after the `!vm.isBranchExecuting()` early return in `executeOpcode`, so unexecuted-branch opcodes are never charged. Confirmed by `TestChargeOpcodeIgnoresDeadBranch`.

2. **Zero limit** — `Validate()` accepts `limit=0`, meaning an admin can disable an opcode entirely via config (`OP_ECPAIRING=0`). This is a reasonable safety valve. Tested implicitly via `TestComputeLimitPairingTripsBeforePerCallCap` (limit=2 trips at 3rd call, so limit=0 would trip at 1st).

3. **No cross-repo breakage** — `solver`, `bancod`, and other consumers only use opcode constants, `NewPacket`, and `ArkadeScriptHash` from the emulator package. None call `script.Execute` or `application.New` directly. The `introspector-review` repo references an older module path and isn't affected.

4. **Test coverage is good** — limit enforcement, dead-branch exemption, override, no-taproot-context bypass, pairing integration, sig counting, and config parsing are all tested. The config tests cover empty/single/multi/unknown/negative/malformed cases.

5. **BIP-342 sigops removal** — Removing the witness-size-proportional budget is correct for Arkade. The L1 budget models block-weight economics that don't apply here. The new flat-50-per-sig limit is more predictable and sufficient for realistic multisig patterns.

---

### ⚠️ Protocol-Critical — Human Review Required

This PR modifies the script VM execution semantics (opcode limiting, sigops budget removal). While the changes are a security improvement (bounding worst-case CPU), they are **protocol-critical** — they change what scripts are valid. Requires human sign-off before merge.

---

**tl;dr:** Fix the shared-map reference (copy in `NewEngine`), improve `WithComputeLimits` godoc, and optionally add a dedicated error code. Otherwise solid work. Requesting changes for the shared-map issue as it's a latent correctness risk in protocol-critical code.

ghost (changes requested):
## Code Review — Opcode Count Compute Limits

Good design. The flat per-opcode count table is simpler and more appropriate for Arkade than the BIP-342 witness-size-proportional sigops budget. The implementation is clean, well-tested, and the config-override path is solid. A few issues to address:

---

### 🔴 MEDIUM — Shared mutable map in `defaultComputeLimits` (`pkg/arkade/compute_limits.go:47`, `pkg/arkade/engine.go:835`)

`NewEngine` assigns the **shared** `defaultComputeLimits` map directly to `vm.limits`:

```go
var defaultComputeLimits = DefaultComputeLimits()  // one map for the process

// engine.go:835
limits: defaultComputeLimits,  // every engine gets the SAME map reference
```

`ComputeLimits` is `map[byte]int` — a reference type. Today no code writes through `vm.limits[op]`, so this is **safe**. But it's fragile: a future change that does `vm.limits[op] = X` on any engine would silently corrupt **every** engine's limits for the rest of the process lifetime.

**Fix:** Either:
- Copy the map in `NewEngine` (safe, tiny cost — 10 entries), or
- Keep the shared reference but add a prominent `// READ-ONLY — do not mutate` comment on both the var and the field, and add a test that verifies `defaultComputeLimits` is not mutated after engine creation.

Copying is the safer choice for protocol-critical code.

---

### 🟡 LOW — `WithComputeLimits` replaces the entire map (`pkg/arkade/script.go:40-43`)

```go
func WithComputeLimits(c ComputeLimits) ExecuteOption {
    return func(engine *Engine) {
        engine.limits = c
    }
}
```

If a caller passes a partial map (e.g., `ComputeLimits{OP_ECPAIRING: 8}`), **all other heavy opcodes become unlimited**. The config parser handles this correctly by starting from `DefaultComputeLimits()` and overriding, but a direct API caller of `WithComputeLimits` could easily make this mistake.

**Fix:** Add a godoc warning:

```go
// WithComputeLimits overrides the per-input opcode-execution compute brake for
// this execution. c replaces the default limits entirely — opcodes absent from
// c become unlimited. Callers that want to override a subset should start from
// DefaultComputeLimits() and modify the returned map.
```

---

### 🟡 LOW — `ErrScriptTooBig` reuse for semantically different error (`pkg/arkade/engine.go:340`)

The compute-limit exhaustion reuses `ErrScriptTooBig`:

```go
return scriptError(txscript.ErrScriptTooBig,
    fmt.Sprintf("opcode %s exceeded execution limit of %d", ...))
```

This is the same error code used at `engine.go:842` for "script size exceeds max allowed size". Callers that match on `ErrScriptTooBig` will conflate "script too large" with "opcode executed too many times." The descriptive string distinguishes them, but error-code consumers won't see the string.

Not blocking, but consider whether a dedicated `ErrComputeLimitExceeded` code would be cleaner for downstream error handling.

---

### 🟢 INFO — Observations (no action needed)

1. **Dead-branch skip is correct** — `chargeOpcode` is called after the `!vm.isBranchExecuting()` early return in `executeOpcode`, so unexecuted-branch opcodes are never charged. Confirmed by `TestChargeOpcodeIgnoresDeadBranch`.

2. **Zero limit** — `Validate()` accepts `limit=0`, meaning an admin can disable an opcode entirely via config (`OP_ECPAIRING=0`). This is a reasonable safety valve. Tested implicitly via `TestComputeLimitPairingTripsBeforePerCallCap` (limit=2 trips at 3rd call, so limit=0 would trip at 1st).

3. **No cross-repo breakage** — `solver`, `bancod`, and other consumers only use opcode constants, `NewPacket`, and `ArkadeScriptHash` from the emulator package. None call `script.Execute` or `application.New` directly. The `introspector-review` repo references an older module path and isn't affected.

4. **Test coverage is good** — limit enforcement, dead-branch exemption, override, no-taproot-context bypass, pairing integration, sig counting, and config parsing are all tested. The config tests cover empty/single/multi/unknown/negative/malformed cases.

5. **BIP-342 sigops removal** — Removing the witness-size-proportional budget is correct for Arkade. The L1 budget models block-weight economics that don't apply here. The new flat-50-per-sig limit is more predictable and sufficient for realistic multisig patterns.

---

### ⚠️ Protocol-Critical — Human Review Required

This PR modifies the script VM execution semantics (opcode limiting, sigops budget removal). While the changes are a security improvement (bounding worst-case CPU), they are **protocol-critical** — they change what scripts are valid. Requires human sign-off before merge.

---

**tl;dr:** Fix the shared-map reference (copy in `NewEngine`), improve `WithComputeLimits` godoc, and optionally add a dedicated error code. Otherwise solid work. Requesting changes for the shared-map issue as it's a latent correctness risk in protocol-critical code.

msinkec:
Note on compute-limit granularity:

The current implementation intentionally uses a very simple `opcode -> max executions per input` lookup table. Limits are independent per opcode, so a worst-case script can spend the full allowance for each configured heavy opcode rather than sharing one grouped budget.

Measured on an Apple M4 Pro, that implemented worst case is roughly **37 ms per input** when all configured heavy-opcode counts are pushed to their limits. A grouped-counter model would bring the same calculation closer to ~24 ms, but the current bound is still far below the multi-second unbounded cases this PR is meant to prevent.

This is a deliberate simplicity tradeoff for now: the table is easy to read, configure, validate, and reason about. If we later need a tighter aggregate CPU bound, we can move from per-opcode counters to grouped counters or a weighted budget without changing the basic charging hook in the engine.

| Opcode / allowance | Measured cost |
|---|---:|
| 50 `OP_CHECKSIG` | 5.27 ms |
| 50 `OP_CHECKSIGVERIFY` | 4.72 ms |
| 50 `OP_CHECKSIGADD` | 4.62 ms |
| 50 `OP_CHECKSIGFROMSTACK` | 4.50 ms |
| 1000 `OP_ECADD` | 2.85 ms |
| 50 `OP_ECMUL` | 2.30 ms |
| 50 `OP_ECMULSCALARVERIFY` | 3.62 ms |
| 50 `OP_TWEAKVERIFY` | 1.63 ms |
| 2 `OP_ECPAIRING` calls, 16 pairs each | 5.05 ms |
| 64 `OP_MODEXP` calls, 64-byte operands | 3.15 ms |
| **Total implemented worst case** | **~37 ms/input** |

ghost (approved):
## Follow-up Review — New Commits `74275cf2`, `0cec6efc`

All three issues from the previous review are resolved:

### ✅ Shared mutable map (was 🔴 MEDIUM)
`NewEngine` now calls `DefaultComputeLimits()` for a fresh copy each time. The `defaultComputeLimits` package-level var is gone. `cloneComputeLimits` helper added for `WithExactComputeLimits`. `TestNewEngineDefaultLimitsAreIndependent` confirms engines don't share state. Clean fix.

### ✅ `WithComputeLimits` partial-map footgun (was 🟡 LOW)
Split into two functions:
- `WithComputeLimits(c)` — merges `c` on top of `DefaultComputeLimits()`, nil is a no-op. Safe for partial overrides.
- `WithExactComputeLimits(c)` — exact replacement with clone. Used by the config path (which already resolved from defaults).

Godoc on both is clear. `TestWithComputeLimitsMergesWithDefaults` and `TestWithExactComputeLimitsCanRemoveDefaultLimit` cover the distinction. `TestWithComputeLimitsNilKeepsDefaultLimits` covers the nil edge case.

### ✅ Config refinements
- Empty value (`OP_ECADD=`) now deletes the limit (makes it unlimited) — tested by `TestParseComputeLimitsEmptyValueRemovesLimit`.
- Empty pair and trailing comma now error — tested.
- Aggregate-cost comment added to `DefaultComputeLimits()`.
- Call sites correctly use `WithExactComputeLimits` since the config path already merges with defaults.

### Data flow verification
`parseComputeLimits("")` → returns full `DefaultComputeLimits()` (never nil) → stored in `Config.ComputeLimits` → passed to `application.New` → `s.computeLimits` → `WithExactComputeLimits(s.computeLimits)`. Correct at every step.

---

⚠️ **Protocol-critical reminder still applies** — this changes VM execution semantics. Human sign-off required before merge.

ghost (approved):
## Follow-up Review — New Commits `74275cf2`, `0cec6efc`

All three issues from the previous review are resolved:

### ✅ Shared mutable map (was 🔴 MEDIUM)
`NewEngine` now calls `DefaultComputeLimits()` for a fresh copy each time. The `defaultComputeLimits` package-level var is gone. `cloneComputeLimits` helper added for `WithExactComputeLimits`. `TestNewEngineDefaultLimitsAreIndependent` confirms engines don't share state. Clean fix.

### ✅ `WithComputeLimits` partial-map footgun (was 🟡 LOW)
Split into two functions:
- `WithComputeLimits(c)` — merges `c` on top of `DefaultComputeLimits()`, nil is a no-op. Safe for partial overrides.
- `WithExactComputeLimits(c)` — exact replacement with clone. Used by the config path (which already resolved from defaults).

Godoc on both is clear. `TestWithComputeLimitsMergesWithDefaults` and `TestWithExactComputeLimitsCanRemoveDefaultLimit` cover the distinction. `TestWithComputeLimitsNilKeepsDefaultLimits` covers the nil edge case.

### ✅ Config refinements
- Empty value (`OP_ECADD=`) now deletes the limit (makes it unlimited) — tested by `TestParseComputeLimitsEmptyValueRemovesLimit`.
- Empty pair and trailing comma now error — tested.
- Aggregate-cost comment added to `DefaultComputeLimits()`.
- Call sites correctly use `WithExactComputeLimits` since the config path already merges with defaults.

### Data flow verification
`parseComputeLimits("")` → returns full `DefaultComputeLimits()` (never nil) → stored in `Config.ComputeLimits` → passed to `application.New` → `s.computeLimits` → `WithExactComputeLimits(s.computeLimits)`. Correct at every step.

---

⚠️ **Protocol-critical reminder still applies** — this changes VM execution semantics. Human sign-off required before merge.

internal/config/config.go
```diff
@@ -23,6 +25,7 @@ const (
 	TLSExtraDomains = "TLS_EXTRA_DOMAINS"
 	LogLevel        = "LOG_LEVEL"
 	ArkdURL         = "ARKD_URL"
+	ComputeLimits   = "COMPUTE_LIMITS"
```
louisinger, line 28: let's update README with description of the COMPUTE_LIMITS format

pkg/arkade/compute_limits.go
```diff
@@ -0,0 +1,60 @@
...
+	return nil
+}
+
+// DefaultComputeLimits returns the canonical per-input execution caps for the
+// heavy opcodes. It returns a fresh map so callers may modify their copy
+// without affecting the engine default.
+//
+// Limits are calibrated from the issue #81 benchmark table (Apple Silicon,
+// single-threaded) to bound worst-case per-input CPU well below the
+// multi-second cases an unbounded script could reach.
```
louisinger:
```suggestion

```

louisinger: let's remove this part of the comment to make it ligther

pkg/arkade/compute_limits_engine_test.go
```diff
@@ -0,0 +1,206 @@
...
+// limitOnlyOp1 builds an engine whose only compute limit is OP_1, with a
+// tapscript execution context active so charges take effect.
+func limitOnlyOp1(t *testing.T, limit int) *Engine {
+	t.Helper()
+	vm, err := newOpcodeEngine(buildOpcodeWorld(), 0)
+	require.NoError(t, err)
+	vm.taprootCtx = &taprootExecutionCtx{}
+	vm.limits = ComputeLimits{OP_1: limit}
+	return vm
+}
```
louisinger: move helper after the Test 🙏

pkg/arkade/compute_limits_test.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func TestDefaultComputeLimitsIsIndependentCopy(t *testing.T) {
+	c1 := DefaultComputeLimits()
+	c1[OP_ECPAIRING] = 99999
+
+	c2 := DefaultComputeLimits()
+	require.NotEqual(t, 99999, c2[OP_ECPAIRING],
+		"mutating a returned copy must not affect later calls")
+}
```
louisinger:
```suggestion
```
let's drop that test to simplify. we can admit default is never mutated by the dev ?

internal/config/config.go
```diff
@@ -84,6 +88,11 @@ func LoadConfig() (*Config, error) {
 	logLevel := viper.GetInt(LogLevel)
 	log.SetLevel(log.Level(logLevel))
 
+	computeLimits, err := parseComputeLimits(viper.GetString(ComputeLimits))
```
louisinger, line 91:
```suggestion
	computeLimits, err := parseComputeLimits(viper.GetStringSlice(ComputeLimits))
```
use GetStringSlice : it will do the comma separated parsing for you so u can simplify `parseComputeLimits`

louisinger (changes requested):

internal/config/config.go
msinkec: This doesn't work for csv env strings.

louisinger: ok ok my bad

louisinger (approved):

chris-ricketts (approved):
Wondering if the `opCounts` should be exposed in the `StepInfo`, as well as an option to count but not error on limit exceeded.

E.g. I could imagine a scenario where someone is building a script and they want to count the limited opcodes in the non-dead paths for debugging/optimisation.

However, this could be in a future PR if deemed necessary.

Otherwise looks solid - LGTM

pkg/arkade/compute_limits_engine_test.go
```diff
@@ -0,0 +1,206 @@
+package arkade
```
louisinger, line 1: @msinkec let's also add a `TestChargeOpcodeZeroLimitForbidsExecution` (or whatever the name) to make sure limit "0" disables the usage

louisinger (changes requested):

msinkec:
> Wondering if the `opCounts` should be exposed in the `StepInfo`, as well as an option to count but not error on limit exceeded.
> 
> E.g. I could imagine a scenario where someone is building a script and they want to count the limited opcodes in the non-dead paths for debugging/optimisation.
> 
> However, this could be in a future PR if deemed necessary.
> 
> Otherwise looks solid - LGTM

Let's do this in a seperate PR should the need arise. In the end a debugging tool that would do this besides other things such as stepping through the execution etc would be ideal.

PR #95, opened 2026-06-02
test: DLEQ proof for blinded-signature key equality

Kukks:
## Summary

Adds self-contained Go tests for a **DLEQ (Discrete Log Equality) proof** — the Cashu/BDHKE construction used to prove that a blinded signature was produced with a known public key, without revealing the secret.

Bob proves the same secret `a` links:
- `A  = a*G` (the public key)
- `C' = a*B'` (the signature over Alice's blinded message `B'`)

```
Bob:   r random; R1 = r*G; R2 = r*B'; e = H(R1,R2,A,C'); s = r + e*a  →  (e, s)
Alice: R1 = s*G - e*A; R2 = s*B' - e*C'; accept iff e == H(R1,R2,A,C')
```

If accepted, the `a` in `A = a*G` equals the `a` in `C' = a*B'`.

## What's included

Two complementary tests, **no production code is modified** — all logic lives in the test files:

### 1. `pkg/arkade/dleq_test.go` — reference prove/verify in Go
Prove/verify helpers plus the test, built on the existing `btcec/v2` point/scalar primitives already used in `tweak.go` (`ScalarBaseMultNonConst`, `ScalarMultNonConst`, `AddNonConst`, `ModNScalar`).

### 2. `pkg/arkade/dleq_script_test.go` — verification inside an Arkade Script
Runs the DLEQ verifier **as an actual Arkade Script** through the engine. The public instance `(G, A, B', C')` is baked into the locking script; the proof `(e, s)` is supplied as the **witness**. The script:

- recomputes `R1 = s*G - e*A` and `R2 = s*B' - e*C'` with `OP_ECMUL`/`OP_ECADD`, using `(n-e)*P` for the subtraction (there is no point-negation opcode);
- reconstructs the SEC1 **compressed** points from their affine `(x, y)` coordinates (`OP_NUM2BIN`/`OP_REVERSEBYTES`/`OP_RIGHT`/`OP_CAT`), which also binds `A`/`C'` between the EC equations and the hash;
- derives the Fiat-Shamir challenge with `OP_SHA256` + `OP_BIN2NUM` + `OP_MOD n`, appending a synthesized `0x00` byte so the little-endian digest is never read as negative;
- asserts the recomputed challenge equals the witness `e`.

## Test cases

| Case | `dleq_test.go` | `dleq_script_test.go` |
|------|:---:|:---:|
| Valid proof | ✅ accept | ✅ accept |
| Wrong `A` | ❌ reject | — |
[...]

PR #94, opened 2026-06-01
Migrate regtest usage off nigiri (in-house Node CLI)

Kukks:
## Summary

`arkade-regtest` was rewritten to remove nigiri / chopsticks / esplora and instead run an in-house Docker Compose stack (Bitcoin Core + Fulcrum + mempool + NBXplorer + arkd + …) driven by a zero-dependency Node CLI (`regtest.mjs`). This PR migrates **this repo's own** standalone integration-test environment off nigiri to match.

The emulator does **not** vendor arkade-regtest as a git submodule and does not use its `start-env.sh`/`regtest.sh` wrappers — it consumes nigiri directly (external `nigiri` Docker network in `docker-compose.regtest.yml`, the `vulpemventures/nigiri-github-action` in CI, and `runCommand("nigiri", …)` calls in the Go integration tests). The changes below reflect that consumption pattern.

### Changes
- **Esplora REST endpoint moved to mempool's `/api`:**
  - Host side: `http://localhost:3000` → `http://localhost:3000/api` — every `mempoolexplorer.NewExplorer(...)` call across `test/*.go` (19 occurrences).
  - In-network: `http://chopsticks:3000` → `http://mempool_web/api` — `ARKD_ESPLORA_URL` in `docker-compose.regtest.yml` and the `ark init --explorer` call (runs inside the `arkd` container).
  - Electrum/Fulcrum endpoints are unchanged (this repo does not use them).
- **Removed direct `nigiri` CLI use in Go integration tests**, replaced with the documented container interface `docker exec bitcoin bitcoin-cli …` via three helpers in `test/tx_test.go`:
  - `onchainFaucet(addr, amountBtc)` — `sendtoaddress` + mine 1 (replaces `nigiri faucet`), returns the funding txid.
  - `mineBlocks(n)` — `bitcoin-cli -generate n` (replaces `nigiri rpc -generate/--generate`).
  - `bitcoinCli(args…)` — generic passthrough (replaces `nigiri rpc …`, e.g. `getrawtransaction`).
- **CI (`.github/workflows/test.yaml`):** replaced the `vulpemventures/nigiri-github-action@v1` step with `actions/setup-node` + cloning arkade-regtest and starting its base profile via the Node CLI (`REGTEST_PROFILES=base node regtest/regtest.mjs start`); added a teardown that runs `regtest.mjs clean`.
- **README:** updated Prerequisites (dropped Nigiri, added Node.js) and the Testing steps to use the arkade-regtest Node CLI.
[...]

PR #96, opened 2026-06-03
feat: Cashu nullifier-pool PoC (trust-minimized ecash exit)

Kukks:
@
## Summary

A **trust-minimized Cashu ecash exit verified entirely in Arkade Script** — the no-SNARK, buildable-today sibling of the `v1-zec` shielded-pool design. A shared fixed-denomination pool VTXO lets any holder of a genuine mint token claim a denomination by proving the token in-script and inserting its nullifier into an on-chain Indexed Merkle Tree, with a recursive covenant rolling the pool state forward. **No mint involvement; no `OP_VERIFY_ZKP`.**

Cashus DLEQ (the mint signature) replaces v1-zecs commitment tree + SNARK — so only the nullifier IMT is needed, on opcodes that exist today.

> Stacked on #95 (reuses its `runArkadeScript`/`encodeBig` harness). Base will retarget to `master` once #95 merges.

## Whats included (all engine-level, fully tested)

**Off-chain reference library** `pkg/arkade/cashupool/` — the byte-for-byte source of truth:
- `hashtocurve.go` — NUT-00 `hash_to_curve` + secret grinder
- `dleq.go` — NUT-12 r-blinded proof DLEQ (uncompressed-point hash, big-endian `e`)
- `imt.go` — depth-parameterized Indexed Merkle Tree (non-membership + coupled insert)
- `state.go` — `PoolState` codec (`prev_root || new_root || size`)

**In-script verifier** (`pkg/arkade/cashupool_*_script_test.go`, package `arkade`):
- `hash_to_curve` with a bounded counter loop (K=4), Euler non-residue via `OP_MODEXP`
- NUT-12 r-blinded DLEQ over `OP_ECMUL`/`OP_ECADD` (the `(n-e)` subtraction trick)
- IMT non-membership + insert via a positional Merkle walk
- **full `ClaimScript`** composing all phases + `0x05` extension-packet state continuity (`OP_INSPECTPACKET`/`OP_INSPECTINPUTPACKET`) + recursive covenant (`OP_INSPECTOUTPUTSCRIPTPUBKEY`) + value conservation — valid claim accepts, **7 negatives reject** (double-spend, tampered e/s, wrong claimant/pool value, broken continuity, wrong new root)

**Engine bugfix** (`5deaa1d`): `OP_CAT` was `append`-ing into aliased stack buffers, corrupting live script bytes when a CAT operand was a data push. Latent today (the existing DLEQ test never CATs a data push); a prerequisite for any data-push-CAT script. Fix allocates a fresh buffer per the documented stack contract.

## Test plan

```
cd pkg/arkade && go test ./... -count=1
```
All `pkg/arkade` and `pkg/arkade/cashupool` tests green; `gofmt`/`go vet` clean.

## Deferred: Phase 6 (live arkd+emulator integration)

The end-to-end integration test (genesis → claim → claim → negatives on a real regtest stack) is **not** in this PR. The repos integration harness is mid-migration off nigiri (#94 → `arkade-regtest#27`, both unmerged), so it cant run green in any currently-available configuration. It will be added against the new Node-driven regtest once that migration lands.

## Design docs (not committed, per repo policy)

`docs/superpowers/specs/2026-06-03-cashu-nullifier-pool-poc-design.md` and the implementation plan capture the full design, decisions, and the v1-zec relationship.
@

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ This is achieved by signing any Arkade transaction (offchain or intent proof) ex
 
 - [`test/htlc_test.go`](test/htlc_test.go) — **Non-interactive HTLC.** A 2-of-2 (`arkd` + emulator-tweaked) VTXO with a claim path gated by HASH160(preimage) and a refund path gated by absolute timelock. Neither the receiver nor the sender ever signs — an arkade covenant enforcing destination + amount replaces both their signatures.
 - [`test/delegate_test.go`](test/delegate_test.go) — **Non-interactive delegate.** A 2-of-2 (`arkd` + emulator-tweaked) VTXO refreshed through batch settlement by any solver, with a CSV exit leaf reserved for the user. The arkade covenant is a self-send (preserves the input's scriptPubKey + value on output 0) gated to intent-proof transactions (`OP_INSPECTVERSION` == 2) so it cannot be drained via off-chain self-send loops.
+- [`pkg/arkade/cashupool_claim_script_test.go`](pkg/arkade/cashupool_claim_script_test.go) — **Trust-minimized Cashu ecash exit.** A shared fixed-denomination pool VTXO whose `ClaimScript` verifies a real Cashu token in-script (NUT-00 `hash_to_curve` + NUT-12 r-blinded DLEQ over the EC opcodes), proves the token's nullifier is unspent and inserts it into an Indexed Merkle Tree (positional Merkle walk), and a recursive covenant rolls the pool state forward via a `0x05` extension packet — no mint involvement. The no-SNARK sibling of the v1-zec shielded-pool design; off-chain reference in [`pkg/arkade/cashupool/`](pkg/arkade/cashupool/).
 
 ## API
 
--- a/pkg/arkade/cashupool/dleq.go
+++ b/pkg/arkade/cashupool/dleq.go
@@ -1,0 +1,112 @@
+package cashupool
+
+import (
+	"crypto/sha256"
+	"math/big"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+)
+
+// --- EC helpers ---
+
+func baseMul(k *btcec.ModNScalar) *btcec.PublicKey {
+	var r btcec.JacobianPoint
+	btcec.ScalarBaseMultNonConst(k, &r)
+	r.ToAffine()
+	return btcec.NewPublicKey(&r.X, &r.Y)
+}
+
+func mul(k *btcec.ModNScalar, p *btcec.PublicKey) *btcec.PublicKey {
+	var pj, r btcec.JacobianPoint
+	p.AsJacobian(&pj)
+	btcec.ScalarMultNonConst(k, &pj, &r)
+	r.ToAffine()
+	return btcec.NewPublicKey(&r.X, &r.Y)
+}
+
+func add(a, b *btcec.PublicKey) *btcec.PublicKey {
+	var aj, bj, r btcec.JacobianPoint
+	a.AsJacobian(&aj)
+	b.AsJacobian(&bj)
+	btcec.AddNonConst(&aj, &bj, &r)
+	r.ToAffine()
+	return btcec.NewPublicKey(&r.X, &r.Y)
+}
+
+// challenge computes e = SHA256(U(R1)||U(R2)||U(A)||U(Cp)) as a big-endian
+// scalar mod n. MUST match the in-script verifier byte-for-byte.
+//
+// Points are fed as uncompressed (65 bytes: 0x04 || X32 || Y32).
+func challenge(R1, R2, A, Cp *btcec.PublicKey) *btcec.ModNScalar {
+	pre := make([]byte, 0, 65*4)
+	pre = append(pre, R1.SerializeUncompressed()...)
+	pre = append(pre, R2.SerializeUncompressed()...)
+	pre = append(pre, A.SerializeUncompressed()...)
+	pre = append(pre, Cp.SerializeUncompressed()...)
+	d := sha256.Sum256(pre)
+	h := new(big.Int).SetBytes(d[:]) // big-endian
+	h.Mod(h, btcec.S256().N)
+	var sc btcec.ModNScalar
+	var b [32]byte
+	h.FillBytes(b[:])
+	sc.SetBytes(&b)
+	return &sc
+}
+
+// Sign is the mint's BDHKE signature C = k*Y.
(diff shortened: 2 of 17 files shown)


ghost (changes requested):
## Arkana Code Review — arkade-os/emulator#96

**Reviewer**: Arkana (aggressive review, protocol-critical)  
**Verdict**: Request changes — **human review required** (exit path, script engine change)

---

### 🔴 PROTOCOL-CRITICAL: OP_CAT buffer aliasing fix (opcode.go:2178)

**Finding**: The fix is **correct and necessary**. Confirmed the full aliasing chain:

1. `opcodePushData` (opcode.go:725) pushes `data` directly — a slice into the script buffer
2. `PickN` (stack.go:357) and `DupN` (stack.go:259) both use `PeekByteArray` which returns references without copying
3. Old code `append(x1, x2...)` writes into x1's spare capacity when `cap(x1) > len(x1)`, corrupting the script buffer or other aliased stack items

The fix allocates a fresh buffer:
```go
result := make([]byte, 0, len(x1)+len(x2))
result = append(result, x1...)
result = append(result, x2...)
```

**This is correct**, but it's a **production engine change** that affects ALL scripts, not just cashupool. It needs:
- [ ] **Dedicated regression test** for the aliasing bug (push literal data, CAT it, verify the script buffer is not corrupted). The PR description says "the existing DLEQ test never CATs a data push" — so there should be an explicit test that _does_, proving the old code would fail and the new code passes.
- [ ] **Audit other append-pattern opcodes** for the same class of bug. Quick scan: `OP_NUM2BIN`, `OP_LEFT`, `OP_RIGHT`, `OP_SUBSTR` — do any of these use `append` on stack items that could alias? This should be checked.

**Flag for human**: This is a latent-to-active correctness fix in the script VM. Even though the fix is obviously correct, a human must sign off on engine changes.

---

### 🟡 Cryptographic Review — DLEQ (cashupool/dleq.go)

**Math is correct.** Verified the Schnorr-like DLEQ proof:
- Prover: `s = p + e*k` where `p` is the deterministic nonce
- Verifier: `R1 = s*G - e*A = p*G` ✓, `R2 = s*B' - e*C' = p*B'` ✓

The uncompressed-point challenge hash and big-endian `e` are consistent between prover, off-chain verifier, and in-script verifier. Good.

**Deterministic nonce** (dleq.go:91): `SHA256("cashupool-dleq-nonce" || k || Y || C)` — correct, prevents nonce reuse. Minor note: this is a custom nonce derivation, not RFC 6979. Acceptable for a PoC but worth documenting the deviation.

---

### 🟡 IMT Non-Membership Range Check (cashupool_claim_script_test.go:1526-1537)

The range check in the claim script:
```
low.Value < k AND (low.Next == 0 OR k < low.Next)
```

**This is correct** but has a subtle edge case worth documenting: if `low.Value == 0` (the genesis sentinel) and `k == 0`, then `low.Value < k` is false, so insertion of nullifier 0 is rejected. This is fine because 0 should never be a valid nullifier (it's the sentinel value), but it should be explicitly documented as an invariant, especially since `Yx` (the nullifier) is a field element that _could_ theoretically be 0 for a pathological token.

**Suggestion**: Add a comment or assertion that `nf > 0` in `prepareClaim`, making the invariant explicit.

---

### 🟡 Covenant Value Conservation (cashupool_claim_script_test.go:1577-1605)

Phase 5 checks:
- `numOutputs == 3`
- `output[0].value == Dsat` (claimant payout)
- `output[1].scriptPubKey == input.scriptPubKey` (recursive covenant)
- `output[1].value == inputValue - Dsat` (pool carries forward)

**This is sound.** Bitcoin consensus prevents output total from exceeding input total, and the covenant locks both the pool value and scriptPubKey. The extension output (index 2) isn't value-checked, but any value there would come from additional inputs and doesn't affect pool integrity.

**One concern**: The script does NOT enforce `numInputs == 1`. An attacker could add a second pool input spending a different pool UTXO. Each would run its own script independently, so this is safe — but worth a comment explaining why multi-input is harmless.

---

### 🟢 Test Coverage Assessment

Strong coverage. The full ClaimScript test exercises:
- ✅ Valid claim (happy path)
- ✅ Double-spend rejection (stale IMT proof vs advanced root)
- ✅ Tampered `e` rejection
- ✅ Wrong claimant value
- ✅ Wrong pool value
- ✅ Broken state continuity
- ✅ Wrong newRoot

**Missing negative test**: tampered `s` in the full claim script (it's tested in the standalone DLEQ script test but not in the composed claim). Not a blocker — the standalone DLEQ test covers it.

**Missing test**: The `prepareClaim` test could be extended to test sequential claims (claim 1 → claim 2 on the advanced pool). The `double-spend` test advances the pool but the second claim doesn't go through legitimately. Consider adding a `chained claims` positive test.

---

### 🟢 Off-chain Reference Library (cashupool/)

Clean, well-tested, no issues:
- `hashtocurve.go`: Matches NUT-00 spec ✓
- `dleq.go`: Matches NUT-12 spec ✓
- `imt.go`: Correct Aztec-style IMT with sorted linked-list threading ✓
- `state.go`: Fixed 68-byte layout with round-trip tests ✓

`findLow` (imt.go:607) iterates all occupied leaves — O(n). `hasOccupiedLeaf` (imt.go:551) scans a range — O(2^level). Both are fine for an off-chain PoC reference but would need indexing for production. Not actionable now.

---

### 🟢 Cross-repo Impact

No downstream consumers found. The `cashupool` package is new and self-contained. The OP_CAT fix is the only change to existing code and has no cross-repo API impact.

---

### Summary

| Finding | Severity | Action |
|---------|----------|--------|
| OP_CAT aliasing fix needs dedicated regression test | 🔴 | Add test proving the bug |
| OP_CAT fix: audit other append-pattern opcodes | 🔴 | Check OP_NUM2BIN, OP_LEFT, OP_RIGHT, OP_SUBSTR |
| Engine change requires human sign-off | 🔴 | Human review |
| Nullifier k=0 invariant undocumented | 🟡 | Add assertion/comment |
| Multi-input safety deserves a comment | 🟡 | Document in covenant section |
| Custom nonce derivation (not RFC 6979) | 🟡 | Document deviation |
| Missing chained-claims positive test | 🟡 | Nice to have |

This is high-quality PoC work — the cryptography is correct, the in-script verification is thorough, and the test coverage is strong. The blocking items are the OP_CAT regression test and the audit of sibling opcodes for the same append-aliasing class.

**Human reviewers**: This PR touches an exit path (ecash claim from a pool VTXO) and the script engine itself. Both are protocol-critical. Please review the OP_CAT fix and the DLEQ/IMT verification logic before merging.

PR #100, opened 2026-06-16
Add missing OP_CAT result size check

msinkec:
## Enforce stack element size limit in OP_CAT

**Problem:** `OP_CAT` concatenated two operands and pushed the result without checking it against `MaxScriptElementSize` (520 bytes). Two valid ≤520-byte elements could produce up to 1040 bytes, and repeated `OP_DUP OP_CAT` doubles the element each round — an unbounded exponential-growth DoS and a consensus-rule violation. It was the only element-producing opcode missing this check (all splice/bitwise/arithmetic/hash/introspection ops are already bounded).

**Fix:** Reject the operation with `ErrElementTooBig` when `len(x1)+len(x2) > MaxScriptElementSize`, matching the pattern already used by the packet-introspection opcodes.

**Tests:**
- Opcode-level: rejects oversized concat (300+300), allows exact boundary (260+260=520).
- Engine-level: `OP_CAT_growth_limit` proves the repeated-doubling attack now fails.

### DIFF
--- a/pkg/arkade/engine_test.go
+++ b/pkg/arkade/engine_test.go
@@ -117,6 +117,38 @@ func TestNewOpcodes(t *testing.T) {
 			},
 		},
 		{
+			// OP_CAT must enforce the per-element size limit
+			// (MaxScriptElementSize == 520 bytes) on its result. Repeatedly
+			// duplicating and concatenating doubles the element size each
+			// round; without the limit this is an exponential-growth DoS.
+			// Starting from a 200-byte element: 200 -> 400 -> 800, so the
+			// second OP_CAT must fail.
+			name: "OP_CAT_growth_limit",
+			script: txscript.NewScriptBuilder().
+				AddOp(OP_DUP).AddOp(OP_CAT). // 200 -> 400
+				AddOp(OP_DUP).AddOp(OP_CAT), // 400 -> 800 (must fail)
+			cases: []testCase{
+				{
+					valid: false,
+					tx: &wire.MsgTx{
+						Version: 1,
+						TxIn: []*wire.TxIn{
+							{
+								PreviousOutPoint: wire.OutPoint{
+									Hash:  chainhash.Hash{},
+									Index: 0,
+								},
+							},
+						},
+					},
+					txIdx:       0,
+					inputAmount: 0,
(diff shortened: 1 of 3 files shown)


ghost (approved):
## ✅ Approved — clean consensus fix

**Verdict:** Approve. This is a correct, well-tested fix for a real consensus violation and DoS vector.

### Code review

**`pkg/arkade/opcode.go:2157-2162` — the fix:**
- Check placement is correct: *before* `append` and *before* `PushByteArray`, so the oversized element never touches the stack.
- Uses `>` (not `>=`), so 520-byte results are correctly allowed. Matches the `MaxScriptElementSize` semantics used everywhere else.
- `len(x1)+len(x2)` cannot overflow: individual elements are already capped at 520 by the engine push-data check (`engine.go:289`), so the sum maxes at 1040 — well within `int` range.
- Pattern exactly matches the existing checks in `opcodeInspectPacket` (`opcode.go:2937`) and `opcodeInspectInputPacket` (`opcode.go:2999`). Consistent.
- No other element-producing opcodes are missing this check: AND/OR/XOR require equal-length operands and produce same-size output; Substr/Left/Right only shrink; BigNum ops go through `PushBigNum` which is bounded by `maxBigNumLen`.

**`pkg/arkade/opcode_test.go` — unit tests:**
- `result_at_max_element_size`: 260+260=520 valid boundary. ✓
- `result_exceeds_max_element_size`: 300+300=600 `ErrElementTooBig`. ✓

**`pkg/arkade/engine_test.go` — integration test:**
- `OP_CAT_growth_limit`: 200→400→800 via `OP_DUP OP_CAT` loop. Proves the exponential-growth DoS path is killed. ✓

**`pkg/arkade/opcode_test.go:1518` — fuzz harness update:**
- `ErrElementTooBig` added to `byteTransformPropertyChecker` accepted errors. Prevents false-positive fuzz failures. ✓

### Cross-repo impact
None. Internal opcode handler change only. No public API, proto, or type changes.

---

⚠️ **Protocol-critical: consensus rule enforcement.** Requires human sign-off before merge.

🤖 Reviewed by Arkana

louisinger (approved):

PR #107, opened 2026-07-02
Add multischeme checksig support

msinkec:
Extends the Arkade Script VM's signature-checking opcodes to verify **ECDSA over secp256k1** and **ECDSA over secp256r1 (NIST P-256)** in addition to the existing BIP340 Schnorr/secp256k1. The scheme is selected by an unambiguous, backwards-compatible extended public-key encoding. This unlocks scripts gated by P-256 signers — passkeys/WebAuthn, Apple Secure Enclave, TPMs, mobile hardware keys — and standard ECDSA keys, without changing anything the emulator itself signs (this is **verification-only**).

Affected opcodes: `OP_CHECKSIG`, `OP_CHECKSIGVERIFY`, `OP_CHECKSIGADD`, `OP_CHECKSIGFROMSTACK`. All four route through a single dispatch layer (`pkg/arkade/sigscheme.go`), so the scheme table lives in exactly one place.

### Public-key encoding

Dispatch is length-first, which keeps existing scripts byte-for-byte compatible:

| Stack pubkey      | Scheme                                    | Key material    |
| ----------------- | ----------------------------------------- | --------------- |
| 32 bytes          | Schnorr / secp256k1 *(legacy, unchanged)* | x-only          |
| `0x10` + 33 bytes | ECDSA / secp256k1                         | SEC1 compressed |
| `0x11` + 33 bytes | ECDSA / secp256r1 (P-256)                 | SEC1 compressed |

Any other prefix or length fails closed with `ErrDiscourageUpgradeablePubKeyType`, preserving BIP342's upgrade-slot semantics (new schemes can be added later without ambiguity). A bare 32-byte key can never collide with an extended key (which is always 34 bytes).

### Signature format & validation

- **64-byte compact `r‖s`** for both algorithms (uniform with the existing Schnorr path). The `OP_CHECKSIG` family still accepts an optional trailing sighash-type byte (65-byte form).
- **ECDSA is non-malleable:** low-s enforced (`s ≤ n/2`) plus canonical scalar bounds (`0 < r,s < n`), checked before the curve operation. Clients normalize DER→compact and high-s→low-s with no signer cooperation.

### Backwards compatibility

- Legacy 32-byte x-only Schnorr/secp256k1 is behaviorally identical (sigCache consult/populate, sighash-type byte handling, nullfail semantics, empty-signature handling all preserved).

### Implementation notes

- **No new dependencies:** secp256k1 ECDSA via `btcec/v2/ecdsa`; P-256 via stdlib `crypto/ecdsa` + `crypto/elliptic`; Schnorr unchanged.
- P-256 verification benchmarks at **~51 µs** (Apple M4 Pro) — actually faster than Schnorr's ~84 µs thanks to Go's native P-256 assembly — so the per-opcode compute-limit cap (50) is unchanged; Schnorr remains the binding cost.


pkg/arkade/sig_multischeme_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"testing"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/stretchr/testify/require"
+)
+
+func checkSigLeafECDSA(t *testing.T, extendedPubKey []byte,
```
louisinger: helpers should be at the end of the file

pkg/arkade/opcode.go
```diff
@@ -2562,35 +2561,23 @@ func opcodeChecksigFromStack(op *opcode, data []byte, vm *Engine) error {
...
 	if err != nil {
 		return err
 	}
 
-	valid := signatureObj.Verify(message, pubKeyObj)
-	if !valid {
-		return scriptError(txscript.ErrNullFail, "schnorr signature verification failed")
+	// The opcode does not hash: message is used exactly as the script placed
+	// it (Schnorr folds it into its BIP340 challenge; ECDSA treats it as the
+	// digest). The script owns any hashing it needs.
```
louisinger:
```suggestion
```

test/csfs_multischeme_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+	"github.com/arkade-os/arkd/pkg/ark-lib/offchain"
+	"github.com/arkade-os/emulator/pkg/arkade"
+	"github.com/btcsuite/btcd/btcutil/psbt"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/stretchr/testify/require"
+)
+
+// Requires the dockerized regtest stack; run through make integrationtest.
```
louisinger:
```suggestion
```
useless comment. `make integrationtest` runs everything inside this folder so yes, it "run through make integrationtest".

pkg/arkade/sigscheme.go
```diff
@@ -0,0 +1,145 @@
...
+		return &schemeKey{scheme: schemeSchnorrSecp256k1, secpPub: pk}, nil
+	}
+
+	prefix := pkBytes[0]
+	key := pkBytes[1:]
+
+	discourage := func() (*schemeKey, error) {
+		return nil, scriptError(txscript.ErrDiscourageUpgradeablePubKeyType,
+			"unsupported pubkey scheme")
+	}
```
louisinger: it's fine to copy paste scriptError twice IMO. no need a custom func for this. if really needed to DRY this code, store the error itself in a variable (no need a func)

louisinger (changes requested):
concept ACK. code quality nACK. please read & clean before commit. very important we start to be consistent in this repo, we'll reach production soon.

msinkec:
Merged sigscheme tests into a single file, I don't think it was a necessary split.

louisinger (approved):

louisinger (approved):

PR #112, opened 2026-07-15
Backoff retry on arkd connect

### LINKED ISSUE #108: handle "server not ready" on startup

louisinger:
This PR extracts the logic of `retryFinalize` to `retryWithBackoff` function. Thus, we can reuse while calling arkd's GetInfo RPC on startup.

it closes #108 

@chris-ricketts please review

## Summary by CodeRabbit

- **Reliability Improvements**
  - Improved startup reliability by automatically retrying connections when the service is not immediately ready.
  - Improved transaction finalization reliability with consistent retry and backoff behavior.
  - Added clearer warning feedback when connection or finalization attempts fail.
  - Retry timing now includes jitter and respects cancellation, helping avoid unnecessary delays.

### DIFF
--- a/internal/application/tx.go
+++ b/internal/application/tx.go
@@ -192,6 +192,15 @@ func (s *service) SubmitTx(ctx context.Context, tx OffchainTx) (*OffchainTx, err
 	}, nil
 }
 
+func (s *service) retryFinalize(ctx context.Context, txid string, checkpoints []string) error {
+	return retryWithBackoff(ctx, finalizeRetryConfig,
+		func() error { return s.arkdClient.FinalizeTx(ctx, txid, checkpoints) },
+		func(attempt int, err error) {
+			log.WithField("txid", txid).WithField("attempt", attempt).Errorf("finalizing tx failed: %s", err)
+		},
+	)
+}
+
 type finalizerAccumulator struct {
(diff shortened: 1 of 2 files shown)


internal/application/tx.go
```diff
@@ -275,47 +284,55 @@ func verifyNonArkdCheckpointSignatures(checkpoints []*psbt.Packet, arkdPubKey *b
...
-
-	for {
-		attempt++
+var arkdConnectRetryConfig = retryConfig{
+	MinAttempts:  0,
+	InitialDelay: 1 * time.Second,
+	MaxDelay:     45 * time.Second,
+	Multiplier:   2.0,
+	Jitter:       0.2,
+}
```
chris-ricketts: While it does not seem worth creating a `utils.go` / `retry.go` file just for the `retryWithBackoff` and `applyJitter` functions, I think this config var should live closer to the function that uses it, i.e. `internal/application/service.go:98-107` - perhaps in-line?

As far as testing goes it seems fine that it `retryWithBackoff`  is covered by `TestRetryFinalize` in `internal/application/tx_test.go:452-513` for both callers.

louisinger: done ae7a38c

chris-ricketts:
LGTM

PR #128, opened 2026-08-06
Reject invalid introspected transaction values

### LINKED ISSUE #124: Reject invalid amounts in value introspection opcodes
## Problem

opcodeInspectInputValue converts the signed int64 previous-output value to uint64 without validating it. A serialized value that decodes to -1 is exposed to the covenant as 18446744073709551615. The fetched previous output is also dereferenced without a nil check.

opcodeInspectOutputValue has the same signed-to-unsigned conversion for transaction outputs.

Relevant code: pkg/arkade/opcode.go.

## Impact

Covenants can evaluate malformed input or output amounts under a different numeric value from the decoded transaction. Exploitation currently composes with permissive sighash types or unauthenticated WitnessUtxo data, but the opcode behavior is unsafe at the library boundary as well.

## Acceptance criteria

- Reject nil previous outputs.
- Reject input and output values outside the valid Bitcoin monetary range before conversion or covenant execution.
- Cover both OP_INSPECTINPUTVALUE and OP_INSPECTOUTPUTVALUE.
- Add focused tests for -1, the maximum valid amount, an amount above the maximum, and a missing previous output.

msinkec:
Closes #124.

Rejects missing previous outputs and transaction values outside Bitcoin's monetary range before either value-introspection opcode converts them to an unsigned script number. Both `OP_INSPECTINPUTVALUE` and `OP_INSPECTOUTPUTVALUE` share the same range check.

Tests cover `-1`, exactly `MaxSatoshi`, `MaxSatoshi + 1`, and a missing previous output.

Validation:
- `make test` from a clean clone
- `go vet ./...` in `pkg/arkade`
- `make integrationtest` against a clean rebuilt regtest stack (`376.148s`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of transaction input and output amounts.
  * Invalid amounts, including negative, oversized, or missing values, are now rejected.
  * Valid amounts up to the maximum supported satoshi value continue to be accepted.

### DIFF
--- a/pkg/arkade/go.mod
+++ b/pkg/arkade/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/arkade-os/arkd/pkg/ark-lib v0.8.1-0.20260318170839-137daaec3a70
 	github.com/btcsuite/btcd v0.24.3-0.20240921052913-67b8efd3ba53
 	github.com/btcsuite/btcd/btcec/v2 v2.3.5
+	github.com/btcsuite/btcd/btcutil v1.1.5
 	github.com/btcsuite/btcd/btcutil/psbt v1.1.9
 	github.com/btcsuite/btcd/chaincfg/chainhash v1.1.0
 	github.com/consensys/gnark-crypto v0.19.2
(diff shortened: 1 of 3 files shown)


chris-ricketts (approved):
LGTM - updated the property checkers and ran FuzzOpcodes for 10 minutes

PR #140, opened 2026-08-19
add OP_TUNNEL for delegated renewal

### LINKED ISSUE #139: Add OP_TUNNEL for exact delegated VTXO renewal
## Summary

Add an Arkade opcode that lets a covenant explicitly authorize an exact one-input/one-output continuation into a fresh Ark batch without an owner signature.

```text
<output_index> OP_TUNNEL -> <true>
```

The opcode is available only through an explicit covenant path. Any compiler-generated default path belongs in the compiler; the emulator must not add an implicit fallback.

## Semantics

For the current real proof input `i` and selected output `k`, `OP_TUNNEL` succeeds only when:

- execution purpose is a decoded `RegisterIntent`, not a transaction version heuristic;
- `i` is an authenticated, unspent off-chain VTXO, excluding the synthetic message input, boarding inputs, unrolled outputs, notes, and on-chain UTXOs;
- the register message has no on-chain output indexes;
- `k` exists and is an ordinary receiver output;
- any Ark extension output is unique and last, preserving proof-to-leaf output indexes;
- the input uses `SIGHASH_ALL` or Taproot default;
- the intent uses arkd's managed tunnel cosigner; and
- neither input `i` nor output `k` has already been used by another successful tunnel invocation in this request.

It then requires exact preservation:

```text
source.scriptPubKey == output[k].scriptPubKey
source.value        == output[k].value
source.assets       == output[k].assets
```

Asset equality compares canonical maps keyed by asset ID. If any execution path invokes `OP_TUNNEL`, the complete intent asset packet must be a pure transfer: every group conserves its total amount and no group performs issuance, reissuance, burn, or metadata operations. This prevents a returned control asset from authorizing new supply elsewhere in the intent.

Application packets remain contract-defined state. The surrounding renewal path must preserve any persistent packets with the existing packet introspection opcodes.

## Expiry gate

With authenticated host values `source_expires_at`, `evaluated_at`, `completion_margin`, and `renewal_window`:

```text
evaluated_at + completion_margin <= source_expires_at
source_expires_at <= evaluated_at + renewal_window
```

Both margins must be non-zero and `completion_margin < renewal_window`. A tunnel register message must also have a non-zero `ExpireAt <= source_expires_at - completion_margin`. Caller-controlled transaction time fields are not substitutes for these values.

Arkd must recheck the completion margin at registration and before round lock, and must assign the successor an expiry at least `completion_margin` later than the source. Those arkd lifecycle checks are host requirements, not opcode behavior.

## VM integration

Tunnel execution needs immutable request context containing:

- execution purpose and decoded register message;
- authenticated source VTXO record, including assets and expiry;
- authorization time and configured margins;
- managed cosigner key; and
- a request-wide input/output claim map shared by all input VMs.

A successful invocation records `(input_index, source_outpoint, output_index)`. Calls in unexecuted branches have no effect, and all recorded state is discarded if script execution fails.

Finalization must verify the existing emulator signature and re-execute the original script and witness with `FinalizationReplay` purpose to recover the actual tunnel mappings. It must not resample time or require the source to remain unspent. Tunnel-authorized inputs may sign only their matching forfeits, never commitment inputs or direct spends.

## Non-goals

Version 1 does not support split, merge, script rotation, value changes, asset transfers, or automatic application-state preservation. It also does not manage tree signing, forfeit collection, or batch failure attribution.

## Acceptance criteria

- Exact BTC and asset continuations succeed without an owner signature.
- Script, value, asset ID/amount, control-asset reissuance, and non-conserving asset changes fail.
- Non-register endpoints, fake inputs, on-chain markings, and non-canonical extension placement fail.
- Expiry boundaries, completion margin, bounded `ExpireAt`, and zero-value settings are covered.
- Duplicate input/output claims fail; independent one-to-one renewals can share an intent.
- An unexecuted `OP_TUNNEL` grants no authorization.
- Finalization reconstructs executed mappings and refuses every commitment-input use of a tunneled outpoint.
- A covenant without an explicit tunnel path receives no delegated authorization.

msinkec:
## Summary

- add `OP_TUNNEL` as an explicit covenant opcode for one-to-one VTXO continuation into a new batch output
- authorize tunnel-bearing register intents against indexed source value, script, assets, status, and renewal timing
- replay the signed authorization during finalization using only proof-committed prevouts and asset data, without adding coordinator/cosigner or stateful completion behavior
- enforce canonical extension placement, reject on-chain outputs, duplicate claims, duplicate signed outpoints, and non-pure or malformed asset transfers
- keep tunnel policy local to emulator configuration and avoid adding a public discovery API

## Why

Delegated renewal should be expressed as an explicit covenant script path rather than an emulator-defined default delegation path. This keeps the opcode general enough for single- and multi-covenant systems while leaving default-path generation to compilers.

## Impact

Covenants can opt into delegated renewal by adding a script branch that invokes `OP_TUNNEL` with the intended successor output index. Existing scripts and emulator behavior remain unchanged unless that opcode is used and the tunnel policy is enabled.

Coordinator/cosigner and watchtower-style batch completion are intentionally out of scope.

## Validation

- `make test`
- `go test ./...` in `pkg/arkade`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added tunnel authorization support for intent registration and finalization.
  * Added configurable tunnel renewal windows and completion margins.
  * Added validation for tunnel sources, assets, timing, and script execution.
  * Added the `OP_TUNNEL` script operation.

* **Bug Fixes**
  * Rejects intent proofs with invalid commitments or duplicate signed inputs.
  * Prevents tunneled inputs from being incorrectly signed as commitment inputs.
  * Preserves and verifies the original encoded intent message.

### DIFF
--- a/api-spec/openapi/swagger/emulator/v1/service.openapi.json
+++ b/api-spec/openapi/swagger/emulator/v1/service.openapi.json
@@ -82,7 +82,7 @@
         "tags": [
           "EmulatorService"
         ],
-        "description": "SubmitIntent signs an intent proof after validating the register message\nand executing Arkade Script on the intent proof transaction.",
+        "description": "SubmitIntent signs an intent proof (any arkd intent message type) after\nvalidating the message and executing Arkade Script on the proof transaction.",
         "operationId": "EmulatorService_SubmitIntent",
         "requestBody": {
           "content": {
@@ -232,6 +232,10 @@
             "type": "string",
             "description": "hex-encoded compressed public key of the signer."
           },
+          "tunnelPolicy": {
+            "$ref": "#/components/schemas/TunnelPolicy",
+            "description": "delegated renewal policy; absent when OP_TUNNEL authorization is disabled."
+          },
           "version": {
             "type": "string"
           }
@@ -373,6 +377,20 @@
             "items": {
               "type": "string"
             }
+          }
+        }
+      },
+      "TunnelPolicy": {
+        "title": "TunnelPolicy",
+        "type": "object",
+        "properties": {
+          "completionMarginSeconds": {
+            "type": "integer",
+            "format": "int64"
+          },
+          "renewalWindowSeconds": {
+            "type": "integer",
+            "format": "int64"
           }
         }
       },
--- a/api-spec/openapi/swagger/emulator/v1/service.openapi.json
+++ b/api-spec/openapi/swagger/emulator/v1/service.openapi.json
@@ -232,10 +232,6 @@
             "type": "string",
             "description": "hex-encoded compressed public key of the signer."
           },
-          "tunnelPolicy": {
-            "$ref": "#/components/schemas/TunnelPolicy",
-            "description": "delegated renewal policy; absent when OP_TUNNEL authorization is disabled."
-          },
           "version": {
             "type": "string"
           }
@@ -377,20 +373,6 @@
             "items": {
               "type": "string"
             }
-          }
-        }
-      },
-      "TunnelPolicy": {
-        "title": "TunnelPolicy",
-        "type": "object",
-        "properties": {
-          "completionMarginSeconds": {
-            "type": "integer",
-            "format": "int64"
-          },
-          "renewalWindowSeconds": {
-            "type": "integer",
-            "format": "int64"
           }
         }
       },
--- a/internal/application/finalization.go
+++ b/internal/application/finalization.go
@@ -21,6 +21,9 @@ import (
 // proof contains a valid emulator signature. Tunnel-bearing scripts are replayed
 // after signature verification to recover their finalization restrictions.
 func (s *service) SubmitFinalization(ctx context.Context, finalization BatchFinalization) (*SignedBatchFinalization, error) {
+	if err := validateIntentMessageCommitment(finalization.Intent); err != nil {
+		return nil, fmt.Errorf("intent message is not committed by the proof: %w", err)
+	}
 	// Sign nothing unless the commitment tx is an arkd-built artifact: a
 	// client-chosen commitment tx (e.g. an ark tx spending a pending
 	// checkpoint outpoint) would let a replayed proof bypass the arkade
@@ -58,7 +61,7 @@ func (s *service) SubmitFinalization(ctx context.Context, finalization BatchFina
 	if len(signedInputs) == 0 {
 		return nil, fmt.Errorf("no signed inputs found in intent proof")
 	}
-	if err := s.replayTunnelAuthorizations(ctx, finalization.Intent, signedInputs); err != nil {
+	if err := s.replayTunnelAuthorizations(finalization.Intent, signedInputs); err != nil {
 		return nil, fmt.Errorf("failed to replay tunnel authorization: %w", err)
 	}
 
@@ -166,11 +169,10 @@ func (s *service) SubmitFinalization(ctx context.Context, finalization BatchFina
 }
 
 type signedInputAssociation struct {
-	script     *arkade.ArkadeScript
-	signer     signer
-	prevout    wire.TxOut
-	inputIndex int
-	tunneled   bool
+	script   *arkade.ArkadeScript
+	signer   signer
+	prevout  wire.TxOut
+	tunneled bool
 }
 
 func (a signedInputAssociation) validateFinalizationInput(
@@ -281,10 +283,13 @@ func getSignedInputAssociations(
 					return nil, fmt.Errorf("invalid signature for input %d", inputIndex)
 				}
 
-				signedInputs[ptx.UnsignedTx.TxIn[inputIndex].PreviousOutPoint] = signedInputAssociation{
-					script:     script,
-					signer:     candidate,
-					inputIndex: inputIndex,
+				outpoint := ptx.UnsignedTx.TxIn[inputIndex].PreviousOutPoint
+				if _, exists := signedInputs[outpoint]; exists {
+					return nil, fmt.Errorf("duplicate signed outpoint %s", outpoint)
+				}
+				signedInputs[outpoint] = signedInputAssociation{
+					script: script,
+					signer: candidate,
 					prevout: wire.TxOut{
 						Value:    input.WitnessUtxo.Value,
 						PkScript: append([]byte(nil), input.WitnessUtxo.PkScript...),
--- a/internal/application/finalization.go
+++ b/internal/application/finalization.go
@@ -17,9 +17,9 @@ import (
 	"github.com/btcsuite/btcd/wire"
 )
 
-// SubmitFinalization doesn't execute arkade scripts, it only signs the forfeits and the commitment tx
-// if and only if the intent proof contains the signer's signature (it means we executed the arkade script in the past)
-// before signing the forfeits, we also verify that is it part of the commitment tx
+// SubmitFinalization signs forfeits and boarding inputs only when the intent
+// proof contains a valid emulator signature. Tunnel-bearing scripts are replayed
+// after signature verification to recover their finalization restrictions.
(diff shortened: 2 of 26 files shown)


Kukks (commented):
Leet's make OP_TUNNEL generic enough to be useful for other things than delegation:

* Allow specifying which things it should enforce tunneling (script, amount, assets, packets)
* vtxo expiry check should be a separate opcode (OP_PUSH_EXPIRY) (should be useful for onchain utxo timelock expiry too)
* opcode to verify we're in an intent (and introsect some of its properties)
*

msinkec:
> * vtxo expiry check should be a separate opcode (OP_PUSH_EXPIRY) (should be useful for onchain utxo timelock expiry too)

Agreed. Could come useful.

> * opcode to verify we're in an intent (and introsect some of its properties)

There's already a spec from Luis for this (OP_INSPECTINTENTMESSSAGE. We could build upon that. 
https://github.com/arkade-os/emulator/issues/68
Also worth mentioning that OP_TUNNEL already implicity enforces that we're spending through an intent tx.

> Leet's make OP_TUNNEL generic enough to be useful for other things than delegation:
> 
> * Allow specifying which things it should enforce tunneling (script, amount, assets, packets)

Not sure about this though. All if these can already be easily restricted with other dedicated opcodes. The purpose of OP_TUNNEL was to bundle requirements for renewal specifically into one opcode for simplicity. Also it provides the lacking spend-through-intent restriction. Provided we would add a seperate OP_PUSHEXPIRY like you proposed and Luis's OP_INSPECTINTENTMESSAGE I don't actually have a reason to have OP_TUNNEL at all apart for cutting a few bytes of the script. I'm planning on abstracting a general pattern of these spending script in the arkade compiler anway, e.g. `contract X implements SimpleDelegate { ... }`

@Kukks

Kukks:
>Also worth mentioning that OP_TUNNEL already implicity enforces that we're spending through an intent tx.

I would remove this

>All if these can already be easily restricted with other dedicated opcodes.

Without loops, asset, packet checks are bounded and very inefficient. If you had a vtxo with 100 assets, you probably cant do it without some obscenity, and youd have to plan for it upfront

msinkec:
Ok I see, so it would be just a sort of copy-all op?
But suppose we actually have a vtxo carrying 100 assets. Let's say it's a dex. Most operations on a dex would change asset state, so copy-all wouldn't suffice. Except for maybe some governance operations and the vtxo renewal itself.
Are you thinking of a more grannual policy which could be passed to the opcode? Something like I proposed earlier, where you could define for example "copy all assets except x". In this case we would need to come up with a policy encoding that's easy to construct in script.

louisinger (changes requested):
* update README to document the new opcode
* we need e2e in `test/delegate_test.go` doing the same "renewal" as TestCovenantDelegate but with OP_TUNNEL

api-spec/openapi/swagger/emulator/v1/service.openapi.json
```diff
@@ -82,7 +82,7 @@
         "tags": [
           "EmulatorService"
         ],
-        "description": "SubmitIntent signs an intent proof after validating the register message\nand executing Arkade Script on the intent proof transaction.",
```
louisinger, line 85: weird ? some problem with proto gen ? 
in any case, this PR doesn't touch the API so revert any changes to api-spec folder

louisinger: re : the comment changed. ok let's move this change out of this PR and open a new one 🙏

internal/application/finalization.go
```diff
@@ -17,10 +17,13 @@ import (
...
 )
 
-// SubmitFinalization doesn't execute arkade scripts, it only signs the forfeits and the commitment tx
-// if and only if the intent proof contains the signer's signature (it means we executed the arkade script in the past)
-// before signing the forfeits, we also verify that is it part of the commitment tx
+// SubmitFinalization signs forfeits and boarding inputs only when the intent
+// proof contains a valid emulator signature. Tunnel-bearing scripts are replayed
+// after signature verification to recover their finalization restrictions.
 func (s *service) SubmitFinalization(ctx context.Context, finalization BatchFinalization) (*SignedBatchFinalization, error) {
+	if err := validateIntentMessageCommitment(finalization.Intent); err != nil {
```
louisinger, line 24: not clear to me why it is necesary for OP_TUNNEL ? it seems a general (good) hardening, maybe worth to cherry pick and move in a dedicated PR ?

api-spec/openapi/swagger/emulator/v1/service.openapi.json
msinkec: Initially the pr included another endpoint to fetch expiry policy but I deferred that and took it out. Seems like a leftover of that.
Anyways, if we're gonna move forward with an OP_PUSHEXPIRY like @Kukks proposed, expiry will be contract defined anyway, so it's not gonna be needed.

internal/application/finalization.go
```diff
@@ -58,6 +61,9 @@ func (s *service) SubmitFinalization(ctx context.Context, finalization BatchFina
 	if len(signedInputs) == 0 {
 		return nil, fmt.Errorf("no signed inputs found in intent proof")
 	}
+	if err := s.replayTunnelAuthorizations(finalization.Intent, signedInputs); err != nil {
+		return nil, fmt.Errorf("failed to replay tunnel authorization: %w", err)
+	}
```
louisinger, line 66: if I understand well, this is to prevent an OP_TUNNEL proof to be used as collab exit. 

ideally I'd like to remove any validation logic. OP_TUNNEL in my mind is just a wrapper for a subset of opcodes (more or less what is done in TestCovenantDelegate).

can we simplify and make OP_TUNNEL behavior as close as the others opcode = without any specific application logic ? OP_TUNNEL should be usable everywhere (onchain tx, offchain, collab exit etc...

I'm also saying this in prevision of future "merge emulator and arkade signer" work (https://github.com/arkade-os/emulator/pull/102). if possible, let's do a first step where we do not touch app layer but only the arkade engine.

pkg/arkade/tunnel.go
```diff
@@ -0,0 +1,228 @@
...
+		return tunnelError(err.Error())
+	}
+	vm.dstack.PushBool(true)
+	return nil
+}
+
+func (c *TunnelContext) authorize(vm *Engine, outputIndex int) error {
+	if c.Purpose != TunnelPurposeRegisterIntent && c.Purpose != TunnelPurposeFinalizationReplay {
+		return fmt.Errorf("invalid execution purpose")
+	}
```
louisinger, line 73: Intent and Finalization are application logic. the engine should never care about it. it seems to me the main tunnel logic can be done via tx output introspections and prevout fetcher. 

is the TunnelContext necessary ? it seems to me extra complexity only to validate some application logic (like for instance disable tunnel if there is onchain outputs etc..). that kind of decisions/validations should live only in the app layer (as much as possible)

msinkec:
Ok, guys let us cut to the chase. I want to confirm we agree on the following:
1. OP_TUNNEL should be generalised to be used as a "bulk" copying op, that enforces data stays the same from the prev VTXO to a new output. We can toggle wether it should copy: scriptPubKey, value, assets (implicitly disables mints / burns, etc.), and other state packets. I guess even these simpler restriction would need some kind of policy encoding in script. I guess a question to debate here is how granual should this policy be, e.g. should we be able to encode a rule saying "preserve all assets in output x, except asset y".
2. To make a spending path enforce that it is for example spendable only via an renwal intent, we implement a seperate OP_INSPECTINTENTMESSAGE.
3. Expiry policy should be handled by a new OP_PUSHEXPIRY.

@Kukks @louisinger

msinkec:
Superseded by a redesigned, general-purpose OP_TUNNEL proposal that separates continuation checks from intent and expiry policy.

PR #141, opened 2026-08-21
Add general OP_TUNNEL output continuation

msinkec:
## Summary

- add a stack-configured OP_TUNNEL opcode for bulk output continuation
- independently preserve the logical VTXO script, bitcoin value, and input-local assets
- allow selected AssetIDs to be excluded for adjacent covenant logic
- keep intent, timing, global asset conservation, and packet policy outside the opcode

## Semantics

OP_TUNNEL is input-local and constrains one selected output. It consumes:

    <output_index> <flags> [<asset_txid> <asset_gidx>]... <exception_count> OP_TUNNEL

The flags independently select exact scriptPubKey (0x01), bitcoin value (0x02), and input-local asset (0x04) continuation. Asset tunneling compares exact AssetID-to-amount maps for the current input and target output, ignoring listed AssetID exceptions on both sides. On success the opcode pushes true; malformed arguments, unavailable source data, or a selected-property mismatch produce a script error.

Script tunneling resolves the logical VTXO behind checkpoints and falls back to the direct prevout for ordinary on-chain spends.

## Why state packets are excluded

Other state packets are transaction-global rather than input- or output-local, so OP_TUNNEL cannot unambiguously associate them with one continuation. There is also no realistic bulk-copy use case for, say, 100 state packets; those are better handled explicitly. A VTXO can plausibly carry that many distinct assets—for example in a DEX—so bulk asset continuation materially reduces covenant size and complexity.

## Testing

- go test ./cmd ./internal/... github.com/arkade-os/emulator/pkg/arkade/... github.com/arkade-os/emulator/pkg/client/...

## Summary by CodeRabbit

* **New Features**
  * Added `OP_TUNNEL` validation to preserve selected output scripts, values, and assets.
  * Supports asset exceptions and local input asset preservation.
  * Validates policies, indexes, flags, and required transaction data.
  * Updated delegate examples to use tunnel validation.

* **Tests**
  * Added comprehensive tunnel validation and fuzz-testing coverage.
  * Expanded pairing validation tests for valid, invalid, boundary, and malformed curve inputs.

* **Documentation**
  * Documented tunnel preservation flags and asset exception arguments.


pkg/arkade/tunnel.go
```diff
@@ -0,0 +1,133 @@
...
+			return tunnelError("selected output does not preserve source value")
+		}
+	}
+	if flags&TunnelAssets != 0 {
+		if err := tunnelAssets(
+			vm.tx.TxHash(), vm.assetPacket, vm.txIdx, int(outputIndex), exceptions,
+		); err != nil {
+			return err
+		}
+	}
```
louisinger, line 96: for value and assets, I think there is an attack vector in case of consolidation.

like I can spend 2 vtxos with the same assets and value and then create a valid intent proof. from the vm point of view, it will be considered valid because the output satisfies both OP_TUNNEL(ed) inputs. so then a malicious actor could "burn" assets.

I think the fix is to enforce the index to be the input's index. we remove <output_index> parameter to the opcode so we disallow consolidation by design via the OP_TUNNEL ?

louisinger (changes requested):
opcode ACK except this possible attack (need to to double check)

we need the delegate e2e test version using OP_TUNNEL to merge

pkg/arkade/tunnel.go
msinkec: That would just be an underconstrained covenant, a dev error. In such a scenerio the script should for example check that it's being spent from a specific input index using OP_PUSHCURRENTINPUTINDEX. You can also easily enforce same in-out position tunneling by using feeding OP_PUSHCURRENTINPUTINDEX as <output_index> to OP_TUNNEL.

PR #147, opened 2026-09-02
Engine: remove verifyWitnessProgram branch

louisinger:
ArkadeScript never runs in the "witness" context. so the whole code `verifyWitnessProgram` was not useful. This PR deletes dead code blocks and simplify the Engine to execute only script in the ArkadeScript context.

@msinkec please review

## Summary by CodeRabbit

- **Changed Behavior**
  - Taproot witness validation has been removed from script execution.
  - Taproot keyspends, tapscript leaf commitments, and annex data are no longer checked during engine execution.
  - Execution errors now consistently evaluate the final stack instead of bypassing checks after a successful keyspend.

- **Tests**
  - Updated Taproot and signature test coverage to use direct witness-stack configuration.

### DIFF
--- a/pkg/arkade/engine.go
+++ b/pkg/arkade/engine.go
@@ -23,11 +23,6 @@ import (
 )
 
 const (
-	// payToTaprootDataSize is the size of the witness program push for
-	// taproot spends. This will be the serialized x-coordinate of the
-	// top-level taproot output public key.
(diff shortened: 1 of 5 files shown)


chris-ricketts (approved):
LGTM - lots of dead code removed

PR #156, opened 2026-09-16
feat: add OP_CHECKTIME

msinkec:
## Summary

- replace OP_CHECKTIMEVERIFY with composable OP_CHECKTIME at opcode 0xdc
- push true for past/current timestamps and false for future timestamps
- preserve the previous behavior with OP_CHECKTIME OP_VERIFY in the delegate covenant

## Tests

- make test
- make integrationtest (PASS, 328.422s)

## Summary by CodeRabbit

* **New Features**
  * Renamed the time-checking opcode to `OP_CHECKTIME`.
  * Time checks now return a boolean result: true when the timestamp has been reached, or false when it is still in the future.
  * Negative timestamps and missing stack values continue to produce errors.

* **Bug Fixes**
  * Future timestamps no longer fail script execution automatically; they can be handled explicitly with verification logic.


chris-ricketts (approved):
LGTM!

louisinger (approved):

PR #157, opened 2026-09-16
test(covenant): add the post-claim reclaim backstop leaf

Kukks:
Adds the post-claim reclaim backstop leaf to the covenant reference. **Stacked on #150** — it extends `test/covenant`, so it should land after that.

## What it adds

`BuildReclaim`, plus `Params.ReclaimLocktime`, a fifth taptree leaf and one validation rule.

The §8.4 fork this resolves: if the receiver never recycles, the operator's carrier sats must not be stranded and the receiver must not lose the asset. `BuildReclaim` recovers the carrier and pins the asset to a hosted receipt for the receiver, behind a long CLTV.

It is the refund split with **one substitution** — vout 1 pays the receiver, not the sender — and that substitution is the entire hazard. Running refund after a claim returns the asset to a sender who has already been paid for it. The two scripts differ by a key rather than by a signature, so nothing but the engine catches a swap. `TestReclaim` asserts both directions refuse the other's payout.

Two properties worth calling out:

- **No sender key on the leaf.** The design makes the signer set itself the hazard, so the post-claim path must not be reachable by the party already paid. `TestReclaimLeafAssembly` asserts the sender's key is absent from the leaf script.
- **Opt-in.** A zero `ReclaimLocktime` omits the leaf entirely, leaving the four-leaf tree and its addresses untouched. Validation requires the backstop to mature after `Locktime`, since maturing first would let the operator reclaim from under a live claim.

## Verification

`go test ./test/covenant/ -count=1` green. Each property mutation-tested and killed by the test that owns it:

| mutation | killed by |
|---|---|
| vout 1 pays the sender | `TestReclaim` |
| sender added to the leaf signer set | `TestReclaimLeafAssembly` |
| locktime ordering check dropped | `rejects_reclaim_locktime_not_after_locktime` |

## This PR was rewritten

It previously added `pkg/arkade/taxi_recycle_test.go` — ~700 lines rebuilding these covenants as test-local helpers. That was a parallel implementation of `test/covenant`, which I had not found because it lives on this unmerged branch rather than on master. It disagreed with the reference on signer sets (it required the claimer, which the covenant already binds via `in[1]`), on dust handling (it modelled only the above-dust branch), and on byte layout.

Every property it asserted is either already covered by #150's own tests — recycle, purchase, refund, asset conservation, params validation, and a `requireRejected` allow-list better than the one I wrote — or has moved into the two tests above. So the file is gone and what remains is the genuinely new leaf.

## One inherited commit

`d4c683a` (Se7enZ, #154) is cherry-picked on top: the grpc v1.83.2 bump for **CVE-2026-84445**. Not mine and not related to the leaf — the base branch pins v1.83.1, and Trivy now fails HIGH on it. The base's own last scan passed only because it ran before this CVE entered the database. Cherry-picked rather than re-authored so the attribution stays correct; it should disappear when #150 rebases on master.

## Not covered

- The **CLTV value** for the backstop. Deliberately a parameter here; the value is a policy decision that is still open.
- Whether the receipt destination is pre-committed at funding or chosen at reclaim time. Also still open, and also mechanism rather than money.
- The TS side. `arkade-taxi`'s `packages/covenant` mirrors this package and its generator pins this repo by commit, so the reclaim leaf needs a `buildReclaim` and a regenerated vector there once this lands.


Kukks:
Valid, and a sharper version of a problem I'd been hitting from the other direction. Addressed in `e2e30cc`.

You're right that the tests couldn't detect incorrect leaf assembly: they built closures inline, serialised, decoded, and asserted the result matched — which proves `ark-lib` round-trips a closure, not that the Taxi leaf set is correct. The test wrote the literals it then checked.

`taxiLeafSet` is now the single place the signer sets are written down, and every assertion consumes it. `postClaimLeaves` is named for the same reason: adding a leaf to the family now forces an explicit decision about whether it's solver-reachable, instead of that decision living implicitly in an inline literal.

Confirmed load-bearing rather than assumed — adding the solver to `reclaimBobOwned` in the shared definition makes the suite report `reclaimBobOwned is solver-reachable: it can revert Bob's asset`.

**Your stronger ask is not done, and I don't want to imply otherwise.** Asserting at the boundary entering `ReadArkadeScript`, or consuming a shared constructor the production path also uses, would mean the test checks what ships rather than a reference written beside it. No such constructor exists in this repo — the contract builders live in the TS SDK, with Go as the reference per `AGENTS.md`, so it lands with the port. The comment on `taxiLeafSet` records that, and the port is required to match this definition rather than copy it.

Worth noting the same class of defect showed up twice more in this file, both found by mutation rather than review, both tests that didn't isolate their property:

- the vout-0 advance equality — the overpay case was rejected by the vout-1 arithmetic instead, so weakening the equality to a floor left the suite green;
- the "Taxi holds no asset" check — routing the asset to Taxi empties vout 1, so the vout-1 lookup rejected first.

Each needed a case constructed so the constraint under test is the only thing that can reject. Your finding is the same shape one level up, at the definition rather than the assertion.

Gate: `go test ./...` in `pkg/arkade` ok, gofmt clean, and all 5 CI jobs were green on the previous head.

Kukks:
## Blocking issue — the finding was correct, and the vectors are gone

`carrier = int64(330)` was wrong. Confirmed independently before responding: `recycleLockup = 330`, `recycleAdvance = 230`, and both reclaim tests use `const carrier = recycleAdvance`. A vector built at 330 pays Taxi the whole lockup and Bob nothing. Exactly as described.

Worth being precise about why my own gate missed it, because it is the same defect this PR is otherwise about. I checked that file by mutation — `advance 230 -> 231` fails the golden comparison — and treated that as sufficient. It is not. A golden that regenerates from the same constant it asserts against only proves the file is self-consistent; it cannot tell you the constant means the right thing. I shipped a test that proves nothing, in the artifact whose entire purpose was to stop that.

The vectors are now reverted outright (`53297b4`), for a reason that subsumes the bug. A Go->TS vector pipeline already exists in the consuming repo: `arkade-taxi` carries `tools/gen-vectors/main.go` and `scripts/gen-vectors.mjs`, and `packages/covenant/test/vectors.json` pins this repo by commit (`49ae96d0`) across 10 cases — recycle, purchase and refund, BTC and asset variants, boundary topups, and both `pinOutput` branches.

That mechanism is strictly better than what I added, and mine set a false bar besides. The Go here models only the above-dust branch; production TS pins a sub-dust payout as an OP_RETURN script matched by hash, orders the `out[0]` value and scriptPubKey checks the other way round, and ends `EQUALVERIFY` plus a pushed `1` rather than `EQUAL`. Same contract, different bytes. "The port must reproduce these exact bytes" would have condemned correct production code.

So the observations about map iteration order and the `claimRecycle`/`refundCooperative` collision are moot — that code is gone. The collision was an artifact of tweaking every leaf over a placeholder `OP_TRUE`; production derives a distinct tweak per covenant, which separates them.

## `assertCovenantRejection` — fixed in `4fde8bd`

Right that a one-string denylist only ever covered the failure that had already bitten. Inverted to an allow-list of the codes a covenant rejection can actually produce (`ErrEvalFalse`, `ErrVerify`, `ErrEqualVerify`); everything else is a harness problem by construction. The original nil-map fixture returns `ErrInvalidIndex`, so that case still fails, and the open-ended rest now fails too.

The guard was itself untested, which is how it shipped incomplete. It is now a pure predicate with direct cases for the three rejections plus stack underflow, malformed push and a non-script error. Verified by mutation: widening the allow-list to `ErrInvalidStackOperation` fails `TestCovenantRejectionGuard/stack_underflow` specifically.

## Two remaining observations

**Two-input pin and fee bumps.** Correct, and deliberate: `INSPECTNUMINPUTS == 2` is what binds `in[1]` to the claimer's own account, which is the constraint that makes the merge safe. Fees come out of the merged output, not a third input. Not adding this to the file comment — per the repo owner's standing rule, design rationale belongs in the PR, not the source.

**`assetConservationBlock` trailing `OP_EQUAL` under composition.** Real, and already handled downstream rather than hypothetical. In `arkade-taxi`'s `packages/covenant/src/scripts.ts`, `buildRecycle` ends its sats block with `EQUALVERIFY` before appending the asset lookups, and a `finish()` helper pushes the trailing `1` only when there is no asset block. So production composes it exactly as you describe it must. These tests execute the blocks standalone, which is why the standalone form is correct here.

## Gate

`go test ./ -count=1` in `pkg/arkade` ok, gofmt clean on the changed file. CI running on `4fde8bd`.

Kukks:
## Correction: the shared constructor does exist

In my earlier reply I answered your "stronger ask" — assert at the boundary, or consume a shared constructor the production path also uses — with:

> No such constructor exists in this repo — the contract builders live in the TS SDK, with Go as the reference per `AGENTS.md`, so it lands with the port.

That is wrong, and I want it corrected on the record rather than left standing.

`test/covenant/covenant.go` exists on `test/dust-free-transfer-covenant` (#150, open), and it is exactly the thing I said was absent: `BuildRecycle`, `BuildPurchase`, `BuildRefund`, `Build`, `PayoutPkScript`, and a `VtxoScript` that assembles the four leaves. It is also what `arkade-taxi`'s vector generator imports, pinning this repo at `49ae96d0`.

I had checked `master`, where `test/covenant` genuinely is absent, and concluded it did not exist anywhere. The reference was on an unmerged branch the whole time.

This matters well beyond the one reply. The helpers in this PR are a parallel reimplementation of that reference, and they disagree with it in ways that favour the reference:

- **Signer sets.** `covenant.VtxoScript` builds leaf 0 as `[server, tweak(recycle)]`. My `taxiLeafSet` has `[claimer, server, tweaked]`. The reference is right — the covenant already pins `in[1]` to the claimer's own account, so requiring the claimer's signature as well is redundant.
- **Dust.** `PayoutPkScript` pins a sub-dust payout as an OP_RETURN script matched by hash. My scripts model the above-dust branch only.
- **Byte layout.** Operand ordering on the `out[0]` checks differs, as does the `EQUALVERIFY`-plus-`1` versus `EQUAL` tail.

So your original observation was the right one and I gave it the wrong answer. The tests here assert properties I still believe are worth asserting, but they assert them against scripts written beside the reference rather than the reference itself — which is the weakness you named in the first pass.

I am not going to quietly rewrite this PR on top of #150 without the repo owner weighing in on sequencing, since #150 is itself open with changes requested. Flagging it here so the duplication is visible to anyone reading this PR in the meantime.

Kukks:
Superseded by #159, which contains every commit here and on this PR's base.

Please keep this branch and `test/dust-free-transfer-covenant` for now: ArkLabsHQ/arkade-taxi pins this repository at `49ae96d` in `scripts/gen-vectors.mjs`, and these two branches are what keep that commit reachable until the pin moves to `master` after #159 merges.

